PR #6, opened 2024-09-25
modified add_theme to color plots with nmfspalette

sbreitbart-NOAA:
In my testing, adding

```
  nmfspalette::scale_color_nmfs() +
  nmfspalette::scale_fill_nmfs() 
```

to the existing plot code worked out fine, regardless of whether the plot is a scatterplot, line plot, etc. But this may fail with other types of plots.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,3 +30,5 @@ Imports:
     stringr,
     tidyr,
     utils
+  Remotes:
+    nmfs-fish-tools/nmfspalette
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,6 +26,7 @@ Imports:
     flextable,
     ggplot2,
     naniar,
+    nmfspalette,
     stringr,
     tidyr,
     utils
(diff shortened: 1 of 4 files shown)


DESCRIPTION
```diff
@@ -26,6 +26,7 @@ Imports:
     flextable,
     ggplot2,
     naniar,
+    nmfspalette,
     stringr,
     tidyr,
     utils
```
Schiano-NOAA, line 32:
```suggestion
    utils
  Remotes:
    nmfs-fish-tools/nmfspalette
```

Schiano-NOAA: @sbreitbart-NOAA you can add remote repositories by adding Remotes: to the description followed by the package path in GitHub. There were some issue with using this in the asar because I added satf as a dependency there, so I am not sure this will 100% work but it should pass the checks. We might need to add that the user needs to download nmfspalette from GitHub in the README

Schiano-NOAA (changes requested):
This looks good now that the remote has been added. Please remove the extrafont package in the DECRIPTION in line 25 and then run devtools::document() in your console then commit those changes. Then go ahead with merging

PR #7, opened 2024-09-26
Generated biomass plot

sbreitbart-NOAA:


R/plot_biomass.R
```diff
@@ -0,0 +1,121 @@
+#' Plot Total Biomass
+#'
+#' @template dat
+#' @template model
```
kellijohnson-NOAA, line 4: The use of `#' @template` is outdated. Guidelines suggest that package developers use `#' @inheritParams`. See guidance here on "[Reusing documentation](https://cran.r-project.org/web/packages/roxygen2/vignettes/reuse.html)".

R/plot_biomass.R
```diff
@@ -0,0 +1,121 @@
+#' Plot Total Biomass
+#'
+#' @template dat
```
Schiano-NOAA, line 3:
```suggestion
#' @inheritparams plot_recruitment
```

R/plot_biomass.R
Schiano-NOAA, line 4:
```suggestion

```

Schiano-NOAA: Thanks for letting us know kelli! I have made the change following the guidance you have provided :)

Issue #11, opened 2024-09-27
`plot_recruitment` cannot inherit params from itself


kellijohnson-NOAA:
The first two arguments of `plot_recruiment()` are not defined in the argument list for roxygen. Instead `@inheritParams plot_recruitment` is listed. This line should be removed and `dat` and `model` should be documented.
https://github.com/nmfs-ost/satf/blob/1641eba94dc762663725213503d450f984bec846/R/plot_recruitment.R#L3

sbreitbart-NOAA:
Looks like this issue was resolved, so I'm going to close it. Please reopen if needed.

Issue #10, opened 2024-09-27
Default should not be hidden


kellijohnson-NOAA:
Instead of using NULL as the default and then checking if it is NULL and then causing something else to happen the user should be aware of what the real default is, which is `"metric ton"` instead of the end result being hidden from them. If the default really is NULL then nothing should be done. Referring to the following hidden change
https://github.com/nmfs-ost/satf/blob/1641eba94dc762663725213503d450f984bec846/R/plot_spawning_biomass.R#L38-L42

Issue #9, opened 2024-09-27
The logic behind `match.arg()` is lost in this `if` `else` statement


kellijohnson-NOAA:
https://github.com/nmfs-ost/satf/blob/1641eba94dc762663725213503d450f984bec846/R/plot_spawning_biomass.R#L30-L34
should be
```
ref_line <- match.arg(ref_line)
```
the first entry of the default values in the function argument is always used used as the default when a user does not specify anything. There is no need for the complications of the if statement or the use of `several.ok = FALSE` because that is the default behavior of `match.arg()`.
See the below example of calling `optim` in R where they also use `match.arg()`
![image](https://github.com/user-attachments/assets/1150a59f-5244-4e1a-91b3-62813e007a98)

PR #13, opened 2024-09-30
Streamline plots

kellijohnson-NOAA:
@Schiano-NOAA this is just a draft PR to show you what I worked on over the weekend. I think that there is a lot of redundant code in the plotting functions that can be worked on. I think the function would be more readable with `switch()` statements instead of nested `if` `else` statements.

Also, I found the following issues.
* There was a problem with `if(any(grepl("target", output$label)))` because it ignored what the user specified in `ref_line` if target is present in label,   i.e., you could never get to MSY. There should be a two-level check, i.e.,   what does the user want and is it present. I have not fixed this latter problem.
* Fixed use of `..., ]$estimate` to `..., "estimate"]` because you should not use the dollar sign which allows for partial matches.
* Fixed the non-relative plots to truncate data to only use years less than end_yr
* Some of the scaled items were hard-coded to 1000 instead of using scale_amount

The following still needs to be fixed:
  * units are not divided by scale amount to create new unit label

I am not sure where you want to go with this. I made some major changes without asking and I wanted to make sure that they are inline with your development plan. If they are not, then maybe we can meet so we can talk about how I can be helpful instead of just changing code that I am not supposed to be changing 😉. Anyway, I wanted to check in before I did anything more.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,6 +24,7 @@ Imports:
     dplyr,
     flextable,
     ggplot2,
+    glue,
     naniar,
     nmfspalette,
     stringr,
--- a/R/plot_biomass.R
+++ b/R/plot_biomass.R
@@ -17,7 +17,6 @@
 #' @export
 #'
 plot_total_biomass <- function(dat,
-                                  model = "standard",
                                   show_warnings = FALSE,
                                   units = NULL,
                                   scaled = FALSE,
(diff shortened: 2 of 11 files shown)


kellijohnson-NOAA:
I can easily pull out the commit with the data and force push the relevant commits back up. If it is okay, I would like to work on this for one more day to double check some things.

Bai-Li-NOAA:
Here are some thoughts on the data: We can use data from the SS3 test models, which I’ve integrated with ASAR. You can modify the code [here](https://github.com/nmfs-ost/asar/blob/main/tests/testthat/fixtures/prepare_ss3_output_files.R) to add some SS3 data to the satf repository for examples and integration tests.

kellijohnson-NOAA:
I redid the PR with a simplified version of plot_spawning_biomass. I tested the figure with two data sets provided by @Schiano-NOAA but there are no examples in the code because there is not a data set within the package to test it with. I could write a test for bogus data if you want. Please see the commit message for items in the PR.

Close #10
Close #11
Close #9 
when merged 😄

Schiano-NOAA:
Leaving a comment because both conflicts and a couple small changes to the code got pushed together:

Fixed some conflicts in the branch and removed the read.csv in favor of output = dat to improve performance for future tasks such as rendering the report per the conversation with @kellijohnson-NOAA. Let me know if you think it should be reverted back to read.csv. 

*Documentation must be updated in response.*

kellijohnson-NOAA:
@Schiano-NOAA I will rebase to master and then force push again. We want to get rid of the four initial commits where I was messing with the data object (i.e., the other pull request). I can change dat to be a read in object though if you want.

Schiano-NOAA:
That sounds good! I think it's probably better to read in as an object rather than the csv.

Schiano-NOAA:
Do it now since you're there anyway. I was planning on doing it after we had figured out how we handle spawning biomass (using it as a model) then apply it to the rest.

kellijohnson-NOAA:
The recent code removes the model argument and changes dat to a read in object rather than a file. I think that this is good to go but you should probably give it a glance or let me know how I can check it with asar before it is merged.

Schiano-NOAA:
Thanks Kelli! I will give it a test on a couple example files then merge this in. As for working with asar, the current connection between the 2 tools is that there are pre-built chunks that contain the satf function already written.

PR #28, opened 2024-11-26
Function that can make rdas for all figures/tables

sbreitbart-NOAA:
This fxn was built with the assumption that all* arguments that are shared among functions (e.g., unit_label), can be assigned a single value. If not, the function should be altered to differentiate the fxns’ identical arguments.

*the exception: ref_line, which was split into ref_line and ref_line_sb


R/exp_all_figs_tables.R
Schiano-NOAA, line 1: Let's discuss this at our meeting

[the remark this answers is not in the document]
Schiano-NOAA: Might need you to walk me through this

Schiano-NOAA (changes requested):
We need to review this together and I need you to show it to me working because I think I am missing something. I could only plot spawning biomass and it never saved the rda anytime I ran it

R/exp_all_figs_tables.R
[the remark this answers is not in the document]
Schiano-NOAA: My folder was not refreshing yesterday so I didn't see anything created - I now see the folder or RDA files so that is good

R/exp_all_figs_tables.R
```diff
@@ -1,44 +1,117 @@
...
+  # figures
+
+  satf::plot_recruitment(dat,
+                         unit_label,
+                         scale_amount,
+                         end_year,
+                         n_projected_years,
+                         relative,
+                         make_rda,
+                         rda_dir)
```
Schiano-NOAA, line 66:
```suggestion
                         rda_dir) |> suppressWarning()
```

sbreitbart-NOAA: Will add to upcoming commit

Schiano-NOAA (approved):
Perfect! Thanks 💯

PR #29, opened 2024-11-27
Extracting numeric (key) quantities

sbreitbart-NOAA:
Wrote code to extract numeric quantities for values associated with biomass and landings plots


R/write_captions.R
```diff
@@ -203,13 +481,78 @@ write_captions <- function(dat, # converted model output object
...
-  # ssb_end_year <- # end year of ssb plot
-  # ssb_units <- # ssb units (plural)
-  # ssb_min <- # minimum ssb
-  # ssb_max <- # maximum ssb
-  # ssb_ref_pt <- # ssb reference point
-  # ssb_ref_pt_units <- # ssb reference point units
+  # start year of ssb plot
+  ssb_start_year <- dat |>
+  dplyr::filter(
+    label == "spawning_biomass",
```
Schiano-NOAA, line 487: does this work for all examples?

R/write_captions.R
```diff
@@ -269,17 +610,274 @@ write_captions <- function(dat, # converted model output object
...
+   # 'landings_tbl_units' = landings_tbl_units,
+   #
+   # ## discards
+   # 'discards_tbl_units' = discards_tbl_units,
+   #
+   # ## catchability
+   # 'catchability_fleet' = catchability_fleet
+  )
+
+  caps_alttext_subbed <- caps_alttext |>
```
Schiano-NOAA, line 867: Can you add a note explaining what this pipeline is doing? I also haven't see the ", ~{" notation before in the below line

Schiano-NOAA (approved):
Looks good overall. I left a couple comments

R/exp_all_figs_tables.R
```diff
@@ -53,6 +52,8 @@ exp_all_figs_tables <- function(
 
 ) {
 
+  make_rda = TRUE
```
Schiano-NOAA, line 55: Why did you take this out of the arguments?

R/write_captions.R
sbreitbart-NOAA: It works for two of my example dat files. Is it failing with yours? Or are there specific files that I should make sure to test it with?

R/write_captions.R
sbreitbart-NOAA: Sure, added the note. The internet helped me with chunk, so here's what I believe is going on: the ~ is present because an anonymous function is passed to another function (dplyr::mutate). The { is separating the for loop from the output of the for loop, which is acted upon by dplyr::mutate.

R/exp_all_figs_tables.R
sbreitbart-NOAA: Because if the user is using this function (to export all figs/tables), that implies that all figures and tables will be exported as rda files. So, make_rda would always be true.

PR #30, opened 2024-12-02
(bug): minor fixes to plots

sbreitbart-NOAA:
Fixes to successfully exclude specific values for year


R/plot_spawn_recruitment.R
```diff
@@ -47,7 +47,7 @@ plot_spawn_recruitment <- function(
                   is.na(sex) | length(unique(sex)) <= 1,
                   is.na(area) | length(unique(area)) <= 1,
                   is.na(growth_pattern) | length(unique(growth_pattern)) <= 1,
-                  year != "S/Rcurve" | year != "Init" | year != "Virg"
+                  !year %in% c('S/Rcurve', 'Init', 'Virg')
```
kellijohnson-NOAA, line 50: I suggest storing this vector of strings as internal data in the package and that way it can be used with multiple functions without having to edit each line in all the files if something changes.

sbreitbart-NOAA: Good idea, Kelli. Just made the change and pushed it, below.

Schiano-NOAA (approved):
Looks good! Thanks for adding in that data object to reference within the functions

PR #34, opened 2024-12-16
Change plot_biomass and spawning_biomass plots to work through errors

Schiano-NOAA:
fix(b and sb plots): reformat plots so they still fxn even missing the reference line

### DIFF
--- a/R/plot_biomass.R
+++ b/R/plot_biomass.R
@@ -8,6 +8,10 @@
 #'   lower-case letters but you must use one of the options specified in the
 #'   default list to ensure that the label on the figure looks correct
 #'   regardless of how it is specified in `dat`.
+#' @param ref_point A known value of the reference point along with the label
+#'   for the reference point as specified in the output file. Please use this
+#'   option if the ref_line cannot find your desired point. Indicate the
+#'   reference point in the form c("label" = value).
 #' @return Plot total biomass from a stock assessment model as found in a NOAA
 #' stock assessment report. Units of total biomass can either be manually added
 #' or will be extracted from the provided file if possible. In later releases, model will not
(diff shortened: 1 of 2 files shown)


sbreitbart-NOAA (commented):
Both fxns seem really flexible and work well under multiple scenarios. I did notice that the spawning biomass plot creates a ref line with an "SB" label, but the biomass plot doesn't have a "B" label. Was that intentional?

Schiano-NOAA:
> Both fxns seem really flexible and work well under multiple scenarios. I did notice that the spawning biomass plot creates a ref line with an "SB" label, but the biomass plot doesn't have a "B" label. Was that intentional?

@sbreitbart-NOAA they are both supposed to have a label when a line is present, but if the value wasn't found in the data set then no line or label should be present. Are you saying that in the example you tested it didn't put a label or that there is no label in the biomass fxn?

Schiano-NOAA:
> > > Both fxns seem really flexible and work well under multiple scenarios. I did notice that the spawning biomass plot creates a ref line with an "SB" label, but the biomass plot doesn't have a "B" label. Was that intentional?
> > 
> > 
> > @sbreitbart-NOAA they are both supposed to have a label when a line is present, but if the value wasn't found in the data set then no line or label should be present. Are you saying that in the example you tested it didn't put a label or that there is no label in the biomass fxn?
> 
> In my 2 example datasets, I can get a ref line to show up, but without a label for the biomass plot

@sbreitbart-NOAA Thanks for catching that! Apparently the function never had any notation through this process. I just updated it

R/plot_biomass.R
```diff
@@ -135,7 +135,15 @@ plot_biomass <- function(
       y = biomass_label) +
     ggplot2::scale_x_continuous(
       n.breaks = x_n_breaks,
-      guide = ggplot2::guide_axis(minor.ticks = TRUE))
+      guide = ggplot2::guide_axis(minor.ticks = TRUE)) +
+    +
```
sbreitbart-NOAA, line 139: Two + signs (lines 138, 139) creates an error

sbreitbart-NOAA (commented):
Besides the extra plus sign, looks good!

PR #37, opened 2024-12-20
Reformat fxns and add new functions

Schiano-NOAA:
Reformat all functions to maintain a standard of implementation as well as introduce a set of new plots and tables. Additional objects will be added in another branch/PR


sbreitbart-NOAA (commented):
My suggestions are in this PR: https://github.com/nmfs-ost/satf/pull/38

PR #40, opened 2024-12-31
Add unit tests

sbreitbart-NOAA:
Making unit tests for `satf`

### DIFF
--- a/R/write_captions.R
+++ b/R/write_captions.R
@@ -622,7 +622,7 @@ write_captions <- function(dat, # converted model output object
     # FIGURES-----
 
    ## kobe plot
-   # 'B.BMSY.min' = as.character(as.character(B.BMSY.min),
+   # 'B.BMSY.min' = as.character(B.BMSY.min),
    # 'B.BMSY.max' = as.character(B.BMSY.max),
    # 'F.FMSY.min' = as.character(F.FMSY.min),
    # 'F.FMSY.max' = as.character(F.FMSY.max),
--- a/R/write_captions.R
+++ b/R/write_captions.R
@@ -28,6 +28,9 @@ write_captions <- function(dat, # converted model output object
   # REMINDERS:
   # -the variable names must exactly match those in the captions/alt text csv.
 
+  # suppress warnings
+  options(warn=-1)
+
   # FIGURES-----
 
   ## kobe plot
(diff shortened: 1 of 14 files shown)


R/write_captions.R
```diff
@@ -615,273 +618,282 @@ write_captions <- function(dat, # converted model output object
...
-   # 'B.BMSY.max' = B.BMSY.max,
-   # 'F.FMSY.min' = F.FMSY.min,
-   # 'F.FMSY.max' = F.FMSY.max,
-   # 'B.BMSY.end.yr' = B.BMSY.end.yr,
-   # 'F.FMSY.end.yr' = F.FMSY.end.yr,
-   # 'overfished.status.is.isnot' = overfished.status.is.isnot,
-   # 'overfishing.status.is.isnot' = overfishing.status.is.isnot,
-   # 'kobe.start.year' = kobe.start.year,
-   # 'kobe.end.year' = kobe.end.year,
+   # 'B.BMSY.min' = as.character(as.character(B.BMSY.min),
```
Schiano-NOAA, line 625: Why are these all set to characters?

tests/testthat/fixtures/sample_data/petrale_sole-after_2020.csv
Schiano-NOAA, line 1: I think this might cause an issue with download times of the package depending on how large this file is. We will need to chat as a group about running a simple SS3 model we can use for a sample data set - for use in both asar and satf

R/write_captions.R
sbreitbart-NOAA: B.BMSY.min should only have one as.character (my oversight with the two- I'll fix that). But the as.character() part is needed in order to replace strings in the df (like B.BMSY.min) with the numeric value of the B.BMSY.min object (like 100, for instance).

tests/testthat/fixtures/sample_data/petrale_sole-after_2020.csv
sbreitbart-NOAA: I filtered the dataset to include data collected in 2020 or later, so it's <5 MB. Is that too large?

Schiano-NOAA (approved):
General regarding questions but looks good otherwise

tests/testthat/fixtures/sample_data/petrale_sole-after_2020.csv
Schiano-NOAA: I think under 10 is fine for a package but we'd have to look up the cran standard. I think though have a simple example would be nice because we would be familiar with it and know what to expect

With that said, let's leave this one in for now and revisit

R/write_captions.R
Schiano-NOAA: Ahh okay makes sense 👍

PR #41, opened 2025-01-02
Failsafe for exp_all_figs_tables

sbreitbart-NOAA:
Add tryCatch to each function that exports a figure or table so that if one function fails, the rest can still produce rdas


Schiano-NOAA (approved):
Looks good!

PR #49, opened 2025-02-21
Added units to plots that can produce relative values

sbreitbart-NOAA:
**NOTE**: After writing the text below, I realized that the issue https://github.com/nmfs-ost/satf/issues/44 was written confusingly (by me 😄). The "relative" argument is not the same thing as the "scaling" argument. I'll just keep this PR going, as a draft, and turn my attention toward adjusting the alt text to adapt to the scaling argument.

---
Specifically edited the alt text for relative biomass, recruitment, and spawning biomass as per https://github.com/nmfs-ost/satf/issues/44.

I think all that's needed to satisfy that issue is to simply state the units (which weren't stated before in these figures' alt texts; only the captions). The key quantities are being computed correctly, and since the alt text _doesn't state the axis units- only the min and max values-_ I believe it's fine to keep using the regular plots' units for these figures.

For example, this is the alt text for relative recruitment:

_Line graph showing the assessment model estimated relative recruitment in sr.units for each year. The x axis shows year, which spans from recruitment.start.year to recruitment.end.year . The y axis shows relative recruitment (R/R~0~), which spans from rel.recruitment.min to rel.recruitment.max **recruitment.units** ._ (I just added the bolded units placeholder in this PR).

If the min and max recruitment are 1,000 and 1,000,000 pounds, then the relative values (R/R0) (could be) 1-1,000 pounds. This alt text would still be sufficient for the relative case:

_"The y axis shows relative recruitment (R/R~0~), which spans from 1 to 1,000 pounds."_


R/exp_all_figs_tables.R
```diff
@@ -3,6 +3,7 @@
 #' Export all figures and tables to Rda files within one function.
 #'
 #' @inheritParams plot_recruitment
+#' @param recruitment_scale_amount The exact amount of scale for recruitment (i.e. 1000)
```
Schiano-NOAA, line 6: I haven't figured out the best way to describe this yet, but I want to add to this something along the lines of indicating that we are scaling down the numbers and this will be reflected in the labels as well

R/exp_all_figs_tables.R
```diff
@@ -15,9 +16,12 @@
 #' for the reference point as specified in the output file. Please use this
 #' option if the ref_line cannot find your desired point. Indicate the
 #' reference point in the form c("label" = value).
+#' @param biomass_scale_amount The exact amount of scale for biomass (i.e. 1000)
```
Schiano-NOAA, line 19: Ditto as above - not sure if you have any ideas

R/exp_all_figs_tables.R
```diff
@@ -15,9 +16,12 @@
...
 #' @param landings_unit_label Units for landings
 #' @param biomass_unit_label Units for biomass
 #' @param spawning_biomass_label Units for spawning biomass
+#' @param spawning_biomass_scale_amount The exact amount of scale for spawning
```
Schiano-NOAA, line 23: Ditto

R/utils.R
```diff
@@ -26,6 +29,13 @@ add_more_key_quants <- function(
     dplyr::filter(label == topic,
                   type == fig_or_table)
 
+  if (!is.null(dat)){
```
Schiano-NOAA, line 32: What happens if it is null? Is there a warning? Would the function just stop?

R/utils.R
```diff
@@ -65,6 +75,153 @@ add_more_key_quants <- function(
...
+      dplyr::filter(label == "biomass",
+                    module_name == "TIME_SERIES" | module_name == "t.series", # SS3 and BAM target module names
+                    is.na(fleet),
+                    is.na(age)) |>
+      dplyr::slice(which.max(estimate)) |>
+      dplyr::select(estimate) |>
+      as.numeric() |>
+      round(digits = 2)
+
+    if (!is.null(scaling)){
```
Schiano-NOAA, line 108: Since scaling is set to 1 as default, this statement can be skipped and you can make the adjustments in the code where I mentioned

R/utils.R
```diff
@@ -65,6 +75,153 @@ add_more_key_quants <- function(
...
+      round(digits = 2)
+
+    # maximum biomass
+    B.max <- dat |>
+      dplyr::filter(label == "biomass",
+                    module_name == "TIME_SERIES" | module_name == "t.series", # SS3 and BAM target module names
+                    is.na(fleet),
+                    is.na(age)) |>
+      dplyr::slice(which.max(estimate)) |>
+      dplyr::select(estimate) |>
```
Schiano-NOAA, line 104:
```suggestion
      dplyr::(estimate / scaling) |>
```

R/utils.R
```diff
@@ -65,6 +75,153 @@ add_more_key_quants <- function(
...
+                    is.na(area) | length(unique(area)) <= 1,
+                    is.na(growth_pattern) | length(unique(growth_pattern)) <= 1,
+                    !year %in% year_exclusions
+      ) |> # SS3 and BAM target module names
+      dplyr::slice(which.max(estimate)) |>
+      dplyr::select(estimate) |>
+      as.numeric() |>
+      round(digits = 2)
+
+    if (!is.null(scaling)){
```
Schiano-NOAA, line 158: See comment and edits above

R/utils.R
```diff
@@ -65,6 +75,153 @@ add_more_key_quants <- function(
...
+  if (topic_cap_alt$label == "biomass") {
+
+    # minimum biomass
+    B.min <- dat |>
+      dplyr::filter(label == "biomass",
+                    module_name == "TIME_SERIES" | module_name == "t.series", # SS3 and BAM target module names
+                    is.na(fleet),
+                    is.na(age)) |>
+      dplyr::slice(which.min(estimate)) |>
+      dplyr::select(estimate) |>
```
Schiano-NOAA, line 93:
```suggestion
      dplyr::pull(estimate / scaling) |>
```

R/utils.R
```diff
@@ -65,6 +75,153 @@ add_more_key_quants <- function(
...
+                    is.na(area) | length(unique(area)) <= 1,
+                    is.na(growth_pattern) | length(unique(growth_pattern)) <= 1,
+                    !year %in% year_exclusions
+      ) |> # SS3 and BAM target module names
+      dplyr::slice(which.max(estimate)) |>
+      dplyr::select(estimate) |>
+      as.numeric() |>
+      round(digits = 2)
+
+    if (!is.null(scaling)){
```
Schiano-NOAA, line 208: See above

R/utils.R
```diff
@@ -82,6 +239,32 @@ add_more_key_quants <- function(
                                                       end_year))
   }
   ## units-----
+
+  if (is.null(scaling)){
```
Schiano-NOAA, line 243: I like this!

Schiano-NOAA (approved):
Looks good! I have some minor suggestions you can adopt or not. I really like how you identified the scaling and how it will change the units

R/exp_all_figs_tables.R
sbreitbart-NOAA: How about something like "A number describing how much to scale down the recruitment quantities shown on the y axis. For example, `recruitment_scale_amount` = 100 would scale down a value from 500,000 --> 5,000"

R/utils.R
sbreitbart-NOAA: It doesn't stop or give a warning. Next commit adds warnings

R/exp_all_figs_tables.R
Schiano-NOAA: @sbreitbart-NOAA I like it!

R/utils.R
sbreitbart-NOAA: Thanks! I bet it can be helpful when adjusting the axis labels too

Schiano-NOAA: @sbreitbart-NOAA for sure!

PR #55, opened 2025-03-05
Create function that makes an html file with all figures and tables

sbreitbart-NOAA:
Function: `html_all_figs_tables()`


inst/WORDLIST
Schiano-NOAA, line 1: I think this function could be reworked so that it's more efficient and does the same thing. When I go to test it, I will see if I can do this

R/html_all_figs_tables.R
Schiano-NOAA, line 1: I think this function could be reworked so that it's more efficient and does the same thing. When I go to test it, I will see if I can do this

tests/testthat/test-html_all_figs_tables.R
```diff
@@ -0,0 +1,104 @@
...
+    ref_line = "unfished",
+    ref_point = 1000,
+    ref_line_sb = "msy",
+    indices_unit_label = "CPUE",
+    rda_dir = getwd()
+  ) |>
+    suppressWarnings()
+
+
+  html_all_figs_tables()
```
Schiano-NOAA, line 22: do you know what the proper convention is for referencing functions from the package in tests? Here you don't have stockplotr:: but in line 7, you do. I think we have some tests in asar set up differently but I keep getting confused about it

tests/testthat/test-html_all_figs_tables.R
Schiano-NOAA, line 1: Nice tests!

R/html_all_figs_tables.R
sbreitbart-NOAA: Sounds good, go for it! Also, since yesterday, I added a missing comma to the DESCRIPTION file so the GH actions errors have changed. Now they're about not being able to find the Quarto PATH. Not sure how to address that yet, but maybe you know?

Schiano-NOAA: I was just about to write back that I reran the actions and it says tests are failing now

tests/testthat/test-html_all_figs_tables.R
sbreitbart-NOAA: I don't know the proper convention. Just did a quick search of Hadley Wickham's tests and see that he doesn't preface the function with the library call. But is there a downside to including the call? I don't think so. So maybe just simple preference?

tests/testthat/test-html_all_figs_tables.R
sbreitbart-NOAA: Thanks!

tests/testthat/test-html_all_figs_tables.R
Schiano-NOAA: Okay good to know! I was just wondering if adding the library call creates a preference for a currently downloaded library rather than the package it's in...if that makes sense. I'm probably just thinking too much into it

sbreitbart-NOAA:
@Schiano-NOAA since your newest commit fixes the failing tests (!), does that mean that our issue is solved and this branch can be merged?

Schiano-NOAA:
@sbreitbart-NOAA Yes! I forgot I reran the checks since they passed and didn't notify me 😆

sbreitbart-NOAA:
> @sbreitbart-NOAA Yes! I forgot I reran the checks since they passed and didn't notify me 😆

Awesome! Thanks so much for finding the solution!

PR #61, opened 2025-03-20
BAA plot

sbreitbart-NOAA:
Draft biomass at age bubble plot.

Note: I removed ref point, ref line, and relative as arguments (compared to the plot_biomass fxn). Are these args needed?

If this looks good, the remaining tasks are:

- [ ] update exp_all_figs_tables to include new fxn
- [ ] make a test for this fxn


R/plot_biomass_at_age.R
```diff
@@ -0,0 +1,162 @@
+#' Plot Total Biomass at Age (BAA)
+#'
+#' @inheritParams plot_recruitment
+#' @return Plot total biomass at age from a stock assessment model as found in a NOAA
+#' stock assessment report. Units of total biomass can either be manually added
+#' or will be extracted from the provided file if possible.
+#' @export
+#'
+plot_biomass_at_age <- function(
+    dat,
```
Schiano-NOAA, line 10: Can you add in an option to scale biomass? See the plot_spawning_biomass has an example on how to implement this

https://github.com/nmfs-ost/stockplotr/blob/f09f63da4f1c94297d2c0bd6f4303b8ed966b581/R/plot_spawning_biomass.R#L113

R/write_captions.R
```diff
@@ -703,6 +703,28 @@ write_captions <- function(dat, # converted model output object
...
+      ) |>
+      dplyr::slice(which.min(age)) |>
+      dplyr::select(age) |>
+      as.numeric()
+
+    # maximum age
+    pop.baa.age.max <- dat |>
+      dplyr::filter(
+        label == "biomass",
+        module_name == "BIOMASS_AT_AGE" | module_name == "B.age", # SS3 and BAM target module names
```
Schiano-NOAA, line 721: Something looks off about the BAM plots so I think we need to go back and check these. The smaller age range is fine but it shouldn't be the same biomass each year

also idk how you might do this atm but the bubbles do not resize with less/more ages. With our atlantic menhaden example, there are only 7 age groups and the bubbles are really small so there is a lot of white space

Schiano-NOAA (changes requested):
This is a very good start. The function works right off the bat. There are minor additions/corrections that I want you to make, but otherwise we can probably add this in soon

R/write_captions.R
sbreitbart-NOAA: 👍 added your first point to our meeting agenda.
And re: the bubbles: we can change the min and max bubble size (currently 0.1 --> 3), but I think the limiting factor is the x axis, which tends to have several decades in it. If we made the max bubble size larger, then we'd have more bubble overlap. If that's fine, then I can up the max bubble size, but I assumed that the user would want to be able to see the distinct bubble edges. Thoughts?

Schiano-NOAA (approved):

PR #63, opened 2025-03-24
Abundance at age plot

sbreitbart-NOAA:
Created function, test, and documentation for new abundance (or numbers) at age bubble plot; added to exp_all_figs_tables fxn


sbreitbart-NOAA:
> This looks good. I think you can either push this through, but if you are able to force the y-axis to maintain whole numbers, that would be a good addition. The menhaden example only has 7 ages and adds ticks in between. This is a small details so nbd if this would be too time consuming. Alternatively, you can add a TODO note inside of the script that we can come back to revisit at some point

Just added this as a new feature and will merge in a moment. I should add this to the biomass at age plot too, right?

Schiano-NOAA:
> > This looks good. I think you can either push this through, but if you are able to force the y-axis to maintain whole numbers, that would be a good addition. The menhaden example only has 7 ages and adds ticks in between. This is a small details so nbd if this would be too time consuming. Alternatively, you can add a TODO note inside of the script that we can come back to revisit at some point
> 
> Just added this as a new feature and will merge in a moment. I should add this to the biomass at age plot too, right?

Yes, please 😄

PR #65, opened 2025-03-27
Update key quantities calculations

sbreitbart-NOAA:
move ssbtarg, ssb.min, ssb.max, rel.ssb.min, and rel.ssb.max from write_captions to add_more_key_quants so that ref_line can be used during calculations when it's added as an argument when running plot_spawning_biomass

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -262,6 +262,9 @@ add_more_key_quants <- function(
           "rel.ssb.max",
           as.character(rel.ssb.max)
         ))
+
+      message(paste0("rel.ssb.min: ", as.character(rel.ssb.min)))
+      message(paste0("rel.ssb.max: ", as.character(rel.ssb.max)))
       }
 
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -175,7 +176,49 @@ add_more_key_quants <- function(
         as.numeric() |>
         round(digits = 2)
 
-      # replace sr.ssb.min and sr.ssb.max placeholders within topic_cap_alt
+      # ssbtarg
+      ssbtarg <- dat |>
```
sbreitbart-NOAA, line 180: @Schiano-NOAA Are you sure this code calculating ssbtarg is correct? I've tried it with multiple dat files and ref_line values and it doesn't seem to work.

sbreitbart-NOAA: Try this, from plot_spawning_biomass:
```dat[
      grep(
        pattern = glue::glue("^spawning_biomass.*{tolower(ref_line)}$"),
        x = dat[["label"]]
      ),
      "estimate"
    ])
```

PR #67, opened 2025-03-27
natural mortality (m) at age plot

sbreitbart-NOAA:
To do:

- [ ]   @Schiano-NOAA check converter- BAM should use a.series, probably, but bsb has only age 0 for that module
- [ ]   Find out how to find M for SS3 models
- [ ]   update M.rate.min, max in write_captions once prev point done
- [ ]   Make unit test
- [ ]   add to exp_all_figs_tables


Schiano-NOAA:
- M is now fixed in the converter for BAM. It will be present in the next asar release.
- M in stock synthesis looks like there an option to have time varying M so its essentially a matrix. For that model, I would just suggest extract M in the last year or when year is NA and it should cover both situations. This plot is only supposed to be for age x M. If a user wants it over time, they can add year as a group or facet once the plot type gets adjusted to our revamped setup

PR #69, opened 2025-03-27
Catch composition plot

sbreitbart-NOAA:
TODO:

 - [x] update alt text/caption to remove fleet name/survey name var?
  -  [ ] update the following dat code to work for BAM (will require work to the converter- Sam will do later)
 -   [x] update the following dat code to work for SS3
 -   [x] update caa vars in write_captions once prev points done
  -  [x] Make unit test
  -  [ ] add to exp_all_figs_tables
  - [x] update asar::create_figures_doc()


R/plot_catch_comp.R
```diff
@@ -0,0 +1,118 @@
...
+  # Units
+  catch_label <- glue::glue("Catch ({unit_label})")
+
+  # TODO:
+  # -update the following dat code to work for BAM
+
+  # read standard data file and extract target quantity
+  if (dim(dat |>
+      dplyr::filter(label == "catch"))[1] > 1) {
+    catch <- dat |>
```
Schiano-NOAA, line 23: Please add a filter that includes end year so the projected catch is not included in the plot - I would suggest adding end_year/year as an argument with the default as the current year

R/plot_catch_comp.R
```diff
@@ -0,0 +1,118 @@
+#' Plot catch composition by fleet
+#'
+#' @inheritParams plot_recruitment
+#' @param unit_label indicate the name of the units of catch as to label the axis
+#'
+#' @return Create a plot ready for a stock assessment report of catch composition
```
Schiano-NOAA, line 6: In this description, please indicate that this is only a plot for when landings are explicitly named in the output file - the current plot does not have the ability to combine all sources of catch - might need steve's input for the wording

sbreitbart-NOAA:
> I'm rethinking about the type of plot for this one. I like the bar graph, but its essentially plotting the same thing as plot_landings if discards aren't modelled separately so maybe we change the plot type to the same as plot_landings. Thoughts?
> 
> I added some minor suggestions for adjustments but the code worked for the necessary examples

Glad it's testing well. Just added updates per your comments.
Re: plot redesign: I think there is a case for making this design identical to the landings plot. If landings != catch, then it would be simple to compare the plots to see similarities and differences in patterns across years. However, if landings = catch, we don't want to produce a redundant plot that doesn't provide any new information. In that case, the user could remove one plot on their own in the figures.qmd.

Schiano-NOAA:
> > I'm rethinking about the type of plot for this one. I like the bar graph, but its essentially plotting the same thing as plot_landings if discards aren't modelled separately so maybe we change the plot type to the same as plot_landings. Thoughts?
> > I added some minor suggestions for adjustments but the code worked for the necessary examples
> 
> Glad it's testing well. Just added updates per your comments. Re: plot redesign: I think there is a case for making this design identical to the landings plot. If landings != catch, then it would be simple to compare the plots to see similarities and differences in patterns across years. However, if landings = catch, we don't want to produce a redundant plot that doesn't provide any new information. In that case, the user could remove one plot on their own in the figures.qmd.

Agreed! I think this all sounds like good considerations and I feel like this plot should be about ready then

sbreitbart-NOAA:
@Schiano-NOAA Just updated the catch plot to match the landings plot (and added some code to landings to match axis formatting of catch). Can you please make sure they're still accurate?

sbreitbart-NOAA:
> @Schiano-NOAA Just updated the catch plot to match the landings plot (and added some code to landings to match axis formatting of catch). Can you please make sure they're still accurate?

ping @Schiano-NOAA

Schiano-NOAA:
> > @Schiano-NOAA Just updated the catch plot to match the landings plot (and added some code to landings to match axis formatting of catch). Can you please make sure they're still accurate?
> 
> ping @Schiano-NOAA

Here are some difference I saw:

- landings are not being filtered by end_year like catch is
- I am wondering why the catch is reported lower than the landings - based on how its calculated - for the sablefish example. I think we should review this with @Steven-Saul-NOAA . This might have to do with the converter and not the plot since this is not the case with the petrale or hake examples
- they have different approaches to the x-axis ticks

Other than this they look good to me! Sorry if this was not what you were asking about when saying "match"

Schiano-NOAA (changes requested):
See my comment for minor changes but otherwise looks good

sbreitbart-NOAA:
> > > @Schiano-NOAA Just updated the catch plot to match the landings plot (and added some code to landings to match axis formatting of catch). Can you please make sure they're still accurate?
> > 
> > 
> > ping @Schiano-NOAA
> 
> Here are some difference I saw:
> 
>     * landings are not being filtered by end_year like catch is
> 
>     * I am wondering why the catch is reported lower than the landings - based on how its calculated - for the sablefish example. I think we should review this with @Steven-Saul-NOAA . This might have to do with the converter and not the plot since this is not the case with the petrale or hake examples
> 
>     * they have different approaches to the x-axis ticks
> 
> 
> Other than this they look good to me! Sorry if this was not what you were asking about when saying "match"

Thanks for reviewing! I think I forgot to review this after a recent meeting and make some changes before pinging you. I'll make these changes and put this question on the agenda for Steve

sbreitbart-NOAA:
Putting this on hold because 1) converter needs updating and 2) catch calculation may need to be tweaked (catch should include selected, retained, and dead biomass)

PR #93, opened 2025-05-01
Adjustments to landings plot

Schiano-NOAA:
Include the argument to indicate fleet names so users that don't have fleets explicit in their output can have it spelled out in the plot


sbreitbart-NOAA (approved):
Looks good and works!

Issue #102, opened 2025-05-09
Move SBtarget to the left


sbreitbart-NOAA:
Currently, SBtarget label may be cut off on some figures, like this:

![Image](https://github.com/user-attachments/assets/92c14dfe-a0c5-489a-a014-7cf56443633d)

PR #104, opened 2025-05-14
Move ref_line labels to avoid getting cut off

sbreitbart-NOAA:
Slight adjustments to the B and SB labels left and down to avoid getting cut off, especially with longer ref line labels like "unfished", as per https://github.com/nmfs-ost/stockplotr/issues/102

### DIFF
--- a/R/plot_biomass.R
+++ b/R/plot_biomass.R
@@ -190,14 +190,12 @@ plot_biomass <- function(
       n.breaks = x_n_breaks,
       guide = ggplot2::guide_axis(minor.ticks = TRUE)
     ) +
-    ggplot2::annotate(
-      geom = "text",
-     # x = as.numeric(end_year) - 3,
-      x = max(b$year)-(0.2*length(b$year)),
-      y = (ref_line_val / ifelse(relative, ref_line_val, scale_amount)) - 1,
+    ggplot2::geom_label(
+      x = max(b$year)-(0.1*length(b$year)),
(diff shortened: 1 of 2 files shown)


R/plot_biomass.R
```diff
@@ -165,8 +165,8 @@ plot_biomass <- function(
     ) +
     ggplot2::annotate(
       geom = "text",
-      x = as.numeric(end_year) + 0.05,
-      y = ref_line_val / ifelse(relative, ref_line_val, scale_amount),
+      x = as.numeric(end_year) - 3,
```
Schiano-NOAA, line 168: This might be a little wonky doing this when values are <1

Schiano-NOAA (commented):
We could add this change in for now, but there could be instances where these changes are too large of a shift. I wonder if it can be a ifelse statement for if the max values is >1 then this is fine otherwise go back to 0.5?

R/plot_biomass.R
Schiano-NOAA: That might work

Schiano-NOAA:
> @Schiano-NOAA I began to make some more changes to the ref_line's appearance but will hold off since we might modularize that feature into its own function.

Do you want me to still do a review on the PR?

sbreitbart-NOAA:
Yes please! Just to ensure that test plots for b and sb look good to you.

Schiano-NOAA (approved):

Issue #79, opened 2025-04-18
Referenced function doesn't have a link


kellijohnson-NOAA:
https://github.com/nmfs-ost/stockplotr/blob/eade4c406d7ac953d7d87bc580a46934685b756f/R/plot_recruitment.R#L3

should be
```
#' @param dat A data frame returned from [asar::convert_output()].
```
to make the function clickable in the documentation. I am guessing that this pertains to many other function references in the documentation as well but I haven't taken the time to find or list them all.

PR #106, opened 2025-05-14
Fix the reference of asar functions in fxn documentation

Schiano-NOAA:
Fix per #79


sbreitbart-NOAA (commented):
When I open up the documentation in the help pane (with `?plot_recruitment`), [asar::convert_output()] is not a hyperlink. Do these changes take effect later on?

Schiano-NOAA:
You know what, I probably need to add (<link>) after the insert

PR #110, opened 2025-06-12
Rename rda folders

sbreitbart-NOAA:
- [x] Do not merge this until this PR has been merged into the main branch of asar: https://github.com/nmfs-ost/asar/pull/280

*When it has:*

- [x] Uncomment tests for html_all_figs_tables; they should work again

This PR:
* ~Replaces 'rda_dir' with 'figures_tables_dir'~
* ~Replaces 'rda_files' with 'figures_tables~
* Replaces 'rda_dir' with 'figures_dir' or 'tables_dir', depending on context
* Replaces 'rda_files' with 'figures' or 'tables', depending on context
* Updates documentation

Motivation: https://github.com/nmfs-ost/workflow_workthrough/discussions/6#discussioncomment-13041875

### DIFF
--- a/R/exp_all_figs_tables.R
+++ b/R/exp_all_figs_tables.R
@@ -60,7 +60,7 @@ exp_all_figs_tables <- function(
     recruitment_scale_amount = 1,
     end_year = NULL,
     relative = FALSE,
-    rda_dir = getwd(),
+    figures_tables_dir = getwd(),
     # imported from plot_biomass
     ref_line = c("target", "MSY", "msy", "unfished"),
     ref_point = NULL,
@@ -106,7 +106,7 @@ exp_all_figs_tables <- function(
         end_year,
         relative,
         make_rda = TRUE,
-        rda_dir
+        figures_tables_dir
       ) #|>
       # suppressWarnings() |>
       # invisible()
@@ -135,7 +135,7 @@ exp_all_figs_tables <- function(
         end_year,
         relative,
         make_rda,
-        rda_dir
+        figures_tables_dir
       ) #|>
       # suppressWarnings() |>
       # invisible()
@@ -161,7 +161,7 @@ exp_all_figs_tables <- function(
         unit_label = landings_unit_label,
         end_year,
         make_rda,
-        rda_dir
+        figures_tables_dir
       ) # |>
       # suppressWarnings() |>
       # invisible()
@@ -182,7 +182,7 @@ exp_all_figs_tables <- function(
         dat,
         end_year,
         make_rda,
-        rda_dir
+        figures_tables_dir
       ) #|>
       # suppressWarnings() |>
       # invisible()
@@ -200,7 +200,7 @@ exp_all_figs_tables <- function(
   #                        recruitment_label = recruitment_unit_label,
   #                        end_year,
   #                        make_rda,
-  #                        rda_dir)# |> suppressWarnings() |> invisible()
+  #                        figures_tables_dir)# |> suppressWarnings() |> invisible()
 
   tryCatch(
     {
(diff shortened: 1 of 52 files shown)


R/exp_all_figs_tables.R
```diff
@@ -60,7 +67,7 @@ exp_all_figs_tables <- function(
     recruitment_scale_amount = 1,
     end_year = NULL,
     relative = FALSE,
-    rda_dir = getwd(),
+    figures_tables_dir = getwd(),
```
Schiano-NOAA, line 70: I think it might be worth just changing this to dir - we could define what this is in the documentation and thinking about it from a R package developer side rather than the specific package at hand, this might be better

R/export_rda.R
```diff
@@ -7,12 +7,11 @@
 #' @param caps_alttext The object containing a figure's caption and alternative
 #' text, in a list, or a table's caption, likely generated with
 #' stockplotr::extract_caps_alttext().
-#' @param rda_dir If the user has already created a folder containing .rda
-#' files with figures, tables, alt text, and captions, rda_dir represents
-#' the location of the folder containing these .rda files ("rda_files").
-#' Otherwise, an "rda_files" folder will be created automatically, then used
-#' to store the exported rda files.
-##' @param topic_label A string that describes a figure or table's label. These
+#' @param figures_tables_dir If the user has already created folders containing
```
Schiano-NOAA, line 10: I would suggest changing this to just dir for same reasoning as i said earlier

R/exp_all_figs_tables.R
sbreitbart-NOAA: dir is already an argument in extract_caps_alttext, utils' add_more_key_quants, and write_captions. I'm hesitant to repeat it, and substitute it in dozens of places, given its ubiquity already. What about a slightly different name, like plot_dir?

R/html_all_figs_tables.R
```diff
@@ -30,22 +42,21 @@ html_all_figs_tables <- function(rda_dir = getwd()) {
...
     asar::create_figures_doc(
-      subdir = tempdir(),
-      include_all = TRUE,
-      rda_dir = rda_dir
+      subdir = figures_tables_dir,
+      figures_dir = figures_tables_dir
     )
 
-    tabs_figs_text <- c(readLines(fs::path(tempdir(), "08_tables.qmd")), readLines(fs::path(tempdir(), "09_figures.qmd")))
+    tabs_figs_text <- c(readLines(fs::path(figures_tables_dir, "08_tables.qmd")),
```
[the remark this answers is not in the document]
sbreitbart-NOAA: Good idea! Just committed the changes.

R/exp_all_figs_tables.R
Schiano-NOAA: Okay no worries!

R/table_landings.R
```diff
@@ -2,6 +2,9 @@
 #'
 #' @inheritParams plot_recruitment
```
Schiano-NOAA, line 3: I'm wondering if it would be worth it to have a plot and a table that contain all of the documentation thats inherited. For example, all tables inherit args from table_landings and all figures inherit args from plot_recruitment

sbreitbart-NOAA: I think that's a great idea. I'm not sure if it'll be possible every time- for example, some argument definitions may need to include words specific to the plot- but I think we should aim for that as much as possible. It'd make the tool less error-prone if we maximized arg inheritance and reduced duplicate definitions

Schiano-NOAA: Yes definitely! Just in the same context as it's used now

PR #115, opened 2025-07-02
Update readme

sbreitbart-NOAA:
-make installation instructions identical to asar
-add example


Schiano-NOAA (commented):
I would add in a Report.sso file in an 'extdata' folder (I think that's the naming convention) but otherwise thanks for updating! Looks good

Issue #112, opened 2025-06-25
Organizing Functions and Major Changes


Schiano-NOAA:
*Preliminary list for review

# General
- [x] @Schiano-NOAA Review unnecessary arguments @ end
- [x] @sbreitbart-NOAA Organize arguments in function based on priority https://github.com/nmfs-ost/stockplotr/issues/81
- [x] Set defaults or required in place of NULL where applicable 
- [x] @sbreitbart-NOAA Substitute cli in interface for messages and such throughout package (Check)
- [x] @Schiano-NOAA Update README
- [x] @sbreitbart-NOAA Add examples to README -- Sam did this when updating the README
- [x] @sbreitbart-NOAA Add more helpful error messages to assist users in debugging/understanding what's going on
- [x] @Schiano-NOAA Look at Ben’s code -- follow up in coworking
- [x] Change end_year default to current year https://github.com/nmfs-ost/stockplotr/pull/122
- [ ] Choose a plot and a table that will contain all of the documentation that every other plot will inherit. For example, all tables inherit args from table_landings and all figures inherit args from plot_recruitment (for end)
    - Plot: @sbreitbart-NOAA suggests plot_biomass or plot_spawning_biomass because they have the most arguments (notably, ref_line and ref_point, which aren't covered in plot_recruitment, which is the plot from which the rest inherit their args)
    - Table: @sbreitbart-NOAA isn't sure, but is leaning towards table_bnc. Rationale: There are only 3 tables, all with the same args except for those associated with units. Indices has no units args; landings has one general arg (unit_label); bnc has three specific args (biomass_unit_label, catch_unit_label, sb_unit_label). It's probably easier for the user to see a simple unit_label when they're making one plot at a time, but if they're using save_all_plots() (which I'm guessing will be used frequently compared to the individual plots), then they'll need arguments with names that differentiate the unit types. In that case, I think we should consider giving each plot's unit-related argument a name that reflects the specific plot so that it can be easily inherited or copied into save_all_plots(). That would mean table_bnc would be the optimal plot here.

# exp_all_figs_tables

@sbreitbart-NOAA https://github.com/nmfs-ost/stockplotr/pull/118

- [x] consider changing function name
- [x] prioritize arguments
- [x] remove any lists and set a default in argument
- [x] set defaults for ALL arguments

# export_rda

@sbreitbart-NOAA https://github.com/nmfs-ost/stockplotr/pull/119

- [x] rename `final` argument to x or object
[...]

PR #118, opened 2025-07-10
Cleanup package: exp_all_figs_tables

sbreitbart-NOAA:
Address items in exp_all_figs_tables section of https://github.com/nmfs-ost/stockplotr/issues/112

**Unresolved questions/comments**:

- [x] What should the default value be for `ref_point` (which applies to several other functions) and `ref_point_sb`?
- [x] I added default values where they were missing. Please double-check these are valid and suggest alternatives where applicable (especially `indices_unit_label`)
- [x] FYI that I didn't change the argument order. I think the existing order works well because it's ordered by plot, so the arguments associated with each plot are grouped together.


R/exp_all_figs_tables.R
```diff
@@ -60,16 +61,16 @@
...
     dat,
     recruitment_unit_label = "mt", # changed from unit_label to recruitment_unit_label for specificity
     recruitment_scale_amount = 1,
-    end_year = NULL,
+    end_year = format(Sys.Date(), "%Y"),
     relative = FALSE,
     figures_tables_dir = getwd(),
     # imported from plot_biomass
-    ref_line = c("target", "MSY", "msy", "unfished"),
+    ref_line = "target",
```
Schiano-NOAA, line 73: I would suggest msy but I have no backing or support on if it's more commonly in an output file than target

R/plot_spawning_biomass.R
```diff
@@ -47,9 +48,9 @@
 #' }
 plot_spawning_biomass <- function(
     dat,
-    unit_label = "metric tons",
+    unit_label = "mt",
     scale_amount = 1,
-    ref_line = c("target", "unfished", "msy"),
+    ref_line = "target",
```
Schiano-NOAA, line 53:
```suggestion
    ref_line = "msy",
```

R/exp_all_figs_tables.R
Schiano-NOAA, line 1: I am assuming the name of this is changed to save_all_plots,. but it doesn't seem to be reflected here

sbreitbart-NOAA: oh yeah good catch!

PR #126, opened 2025-07-31
Streamline tests

sbreitbart-NOAA:
* add sample converted model output in .rda format to fixtures
* update tests so that each imports the .rda instead of converting the output file with asar every time


Schiano-NOAA (commented):
Looks good! I am not sure it's worth the effort/don't know if we should be doing it, but there are references to the package in all of the tests (stockplotr::fxn()). I wonder if they're necessary or not

PR #128, opened 2025-08-06
Cleanup pkg

sbreitbart-NOAA:
Branch with nearly all tasks outlined in https://github.com/nmfs-ost/stockplotr/issues/112

Remaining tasks are related to verifying key quantities and choosing figure/table from which all other plots will inherit documentation (which can be resolved once `revamped-baseline-plot` branch is more developed)

### DIFF
--- a/R/plot_spawning_biomass.R
+++ b/R/plot_spawning_biomass.R
@@ -60,9 +60,7 @@ plot_spawning_biomass <- function(
     ref_line <- names(ref_point)
   } else if (length(ref_line) > 1) {
     ref_line <- "target"
-  } else {
-    ref_line <- ref_line
-  }
+  } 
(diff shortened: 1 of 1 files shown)


R/plot_spawning_biomass.R
```diff
@@ -60,7 +61,7 @@ plot_spawning_biomass <- function(
   } else if (length(ref_line) > 1) {
     ref_line <- "target"
   } else {
```
Schiano-NOAA, line 63: This else statement is no longer needed since there is a default for ref_line. Please remove

Schiano-NOAA (commented):
One minor suggested change but otherwise a great revision of the package (at least until we have the revamped plots done 😄 )

PR #131, opened 2025-08-11
Revamped N-at-age plot

Schiano-NOAA:
**MERGE INTO REVAMPED-SR BEFORE REVAMPED-SR MERGE**

Revamped the abundnace-at-age plot to fit new formatting and modular design


sbreitbart-NOAA:
Rather than submit all my thoughts in one review, I'm going to add some comments in a few parts in case I'm not using these functions correctly or fixes could affect other functions.

First: I started off testing `plot_abundance_at_age(am)` and `plot_abundance_at_age(bsb)`. Both give this error:

Error in `dplyr::summarise()` at stockplotr/R/plot_abundance_at_age.R:118:3:
ℹ In argument: `interm = sum(age *
  years_per_year)`.
ℹ In group 1: `year = 1978`.
Caused by error in `age * years_per_year`:
! non-numeric argument to binary operator

`plot_abundance_at_age(hake)` produces
Error in `dplyr::filter()` at stockplotr/R/plot_abundance_at_age.R:91:5:
ℹ In argument: `time%%1 != 0.5`.
Caused by error in `time %% 1`:
! non-numeric argument to binary operator

R/plot_recruitment_deviations.R
```diff
@@ -1,6 +1,9 @@
 #' Plot Recruitment Deviations
 #'
 #' @inheritParams plot_spawning_biomass
+#' @param group string of a column that groups the data (e.g. "fleet", "area", etc.)
+#' @param facet string or list of strings that facet the data (e.g. "year", "area", etc.)
+#' @param ... Other ggplot2 options for selected geom (e.g. `size`, `color`, etc.)
```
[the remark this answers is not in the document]
Schiano-NOAA: Please ignore the rec devs plot here and test in  revamped-sr. I made changed to it in that branch and that was where it was initially made. Just review the AA plots please

R/plot_abundance_at_age.R
```diff
@@ -30,20 +30,22 @@
 #'
 #' plot_abundance_at_age(
 #'   dat,
-#'   unit_label = "my_unit",
-#'   scale_amount = 10,
-#'   end_year = 2024,
-#'   make_rda = TRUE,
+#'   group = "area",
```
sbreitbart-NOAA, line 33: When there are many facets, the axis labels can get crunched, like here:

 `plot_abundance_at_age(bsb, group = "age")`

Image

What if there was a feature that decreased axis ticks if there are, e.g., >4 faceted plots?

R/plot_abundance_at_age.R
```diff
@@ -30,20 +30,22 @@
...
+#'   unit_label = "fish",
```
sbreitbart-NOAA, line 34: `unit_label` doesn't seem to work properly; did you remove the legend? I think it's important for the reader to see. Ex:

`plot_abundance_at_age(bsb, unit_label = "test unit")`

Image

R/plot_abundance_at_age.R
Schiano-NOAA: I am not sure, but you shouldn't be faceting by age in this plot because the y axis should be age. I think this is still relevant for plots with many fleets. The user could also use ggsave and fixing the dimensions will fix this.

I think it's a generally busy plot and maybe we recommend not facetting?

R/plot_abundance_at_age.R
```diff
@@ -30,20 +30,22 @@
...
+#'   scale_amount = 1000,
```
sbreitbart-NOAA, line 35: Can't test this without the legend

R/plot_abundance_at_age.R
Schiano-NOAA: There should be an argument called "proportional" which changes how the data is handled in the plot. When proportional is TRUE, then the size of the circles in proportion to the abundance. If it's FALSE then its relative to abundance and then will require a legend. 

It looks like I forgot to add it to the documentation, but it's still an argument in the function.

R/plot_abundance_at_age.R
sbreitbart-NOAA: When I test this fxn out with group = "fleet", it only provides one plot with "NA" as the title. Maybe this is expected given its current state and will be resolved once "facet" is added?

If faceting isn't recommended, then that's important to add to the documentation. I think the way we move forward really depends on what kinds of plots SA scientists will want to produce. The ggsave option could work, but if many scientists will have to manually save those plots, then I think automating it is better in the long run. Not sure!

R/plot_abundance_at_age.R
sbreitbart-NOAA: I think we should be clear about the unit of measure, either in the caption (and alt text), or in a legend

R/plot_abundance_at_age.R
Schiano-NOAA: I will check this again and get back to you + fix the documentation and arguments. 

To clarify, faceting **is** recommended.

Also I can definitely make the changes so it's less crowded but when recommending ggsave, it was one way, but the way people view and save these plots will vary base on the program they use to code as well. For example, vscode extension to view plots allows the user to scale everything as they view it. For Rstudio, the view of the plot changes a lot based on the size of your screen and how large you have the plot open

sbreitbart-NOAA:
What do you mean by "all (relevant) tests are passing"?

sbreitbart-NOAA:
There are two more notes that should be pretty quick to address, then let's merge!

1. I think there should be a message if the user tries to run `plot_catch_comp()` on a BAM model, explaining that this function doesn't work for that model output. You could use the same verbiage as in `convert_output()` if the user chooses an incompatible `model` type.
2. There isn't much contrast between the lines of best fit and the bubbles in the `plot_catch_comp()` plot. What do you think about making the lines more visible by making them, e.g., black, but then a bit thinner?
image
3. In the `plot_catch_comp()` plot's alt text, there's a sentence that's not present in the NAA or BAA plots' alt text: "The size of the bubbles range from tot.catch.min to tot.catch.max." I think we should be consistent so that if this sentence is present for one AA plot, it's in the others as well.

4. Lastly, all the key quantities for these plots extracted properly 🥳 except for tot.catch.min to tot.catch.max. I edited utils_rda.R so that these values are rounded and subbed into the alt text csv (will push that commit in a moment)

Schiano-NOAA:
@sbreitbart-NOAA All great additions! I will get those done as soon as I can. Thanks Sophie :)

Issue #107, opened 2025-05-20
Change default color palette


sbreitbart-NOAA:
* Change add_theme default palette from nmfspalette to viridis to increase contrast between colors
* add documentation to vignette explaining to how to alter color palettes
* Update associated test and DESCRIPTION file once nmfspalette is removed

NOTE: Sophie attempted this and is delaying this because there are more pressing issues to address and this will be tricky- mainly, coding add_theme so that the correct palette is applied (discrete or continuous).

PR #135, opened 2025-08-19
Apply viridis palette to figures

sbreitbart-NOAA:
Apply viridis palette to figures, rather than nmfspalette(), for increased accessibility

https://github.com/nmfs-ost/stockplotr/issues/107


Schiano-NOAA (approved):
Woop!

PR #136, opened 2025-08-28
Revamped b plot

sbreitbart-NOAA:
* Revamped biomass plot

@Schiano-NOAA Can you please test this with sablefish and skate? These datasets aren't plotting (at least in the amount of time I'm waiting).


sbreitbart-NOAA:
@Schiano-NOAA should we keep this PR open until the reference line issue is resolved (elsewhere)? And could you please check that the `relative` arg is working as expected?

PR #137, opened 2025-08-29
Revamped recruitment plot

sbreitbart-NOAA:
Revamped recruitment plot

**Question** for @Schiano-NOAA : what's an example of `module`? I haven't seen it in other functions' examples and think we should show some possibilities

### DIFF
--- a/R/plot_recruitment.R
+++ b/R/plot_recruitment.R
@@ -93,7 +93,8 @@ plot_recruitment <- function(
       ggplot2::geom_line(
         data = recruitment,
         ggplot2::aes(x = year, y = expected_recruitment),
-        color = "red"
+        color = "red",
+        size = 1
       )
   }
   
--- a/R/plot_recruitment.R
+++ b/R/plot_recruitment.R
@@ -19,6 +19,7 @@
 #'   scale_amount = 100,
 #'   relative = TRUE,
 #'   interactive = TRUE,
+#'   module = "TIME_SERIES",
 #'   make_rda = TRUE,
 #'   figures_dir = getwd()
 #' )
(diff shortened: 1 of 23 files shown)


Schiano-NOAA:
module = "TIME_SERIES". It would be the selected string if you were to run it interactively for module_name. I just shortened it in the function, but maybe we should change these args to module_name for consistency

Schiano-NOAA:
I am currently reviewing this PR. There are some changes that  need to be made. I think what we want now is a plot more similar to plot_error (no fault of yours). I was looking at the examples in the slidedeck and I think we want recruitment over time with error bars. I am making these adjustments and some changes to plot_error then will push. Then, I'll leave my review with probably some requested changes. Please don't rebase until I put in my review :)

Thanks for your hard work on this!

Schiano-NOAA (changes requested):
I think this is mostly good. I think the line for predicted recruitment should be thicker. I will be adding the commas option into the theming so no need for that.

You will see that I pushed a commit that replaced the plot_timeseries function in plot_recruitment for plot_error as I think it would be more beneficial to see the trend in R with their error when present (more similar to recruitment deviations). I also adjusted the plot_error function because not all uses of that will want the reference line at 1, so I added `hline` to turn that on and off.

Other than that, I think this is basically ready but needs more testing. I found an error in the converter for exporting the error values for this plot, so the fix will be pushed with everything else and you could formally test that.

I am recommending to make the line thicker in this PR then merge into revamped-sr. We can do final adjustments if needed when the converter updates are live. The last thing to do is to resoled the merge conflicts from the revamped-b-plot merge into revamped-sr

PR #145, opened 2025-10-10
Revamped plots round2

Schiano-NOAA:
Revamping the following plots to match recent revamped plots in v0.5.0-beta:

- [x] plot_natural_mortality
- [x] plot_fishing_mortality
- [x] plot_indicies
- [x] plot_landings
- [x] plot_recruitement (fixed issue where model is not on same plot for comparisons)


R/plot_fishing_mortality.R
```diff
@@ -0,0 +1,103 @@
...
+#' @return Create a plot ready for a stock assessment report of fishing mortality over time 
+#' from the results of an assessment model translated to the a standardized output 
+#' (\link[asar]{convert_output}). Includes options to group or facet data where available.
+#' There are options to return a
+#' [ggplot2::ggplot()] object or export an rda object containing associated
+#' caption and alternative text for the figure.
+#' @export
+#'
+#' @examples
+#' plot_fishing_mortality(
```
sbreitbart-NOAA: I ran this example, except I set `group = "none"`, and the output still shows faceting:

```
plot_fishing_mortality(
   dat = stockplotr:::example_data,
   ref_line = c("target" = 0.2),
   group = "none",
   interactive = FALSE
 )
```

Image

R/plot_landings.R
```diff
@@ -1,180 +1,103 @@
...
+#'
+#' @examples
+#' plot_landings(
+#'   dat = stockplotr:::example_data,
+#'   unit_label = "metric tons",
+#'   group = "fleet",
+#'   interactive = FALSE,
+#'   make_rda = FALSE
+#' )
+plot_landings <- function(
```
sbreitbart-NOAA: The linewidth for this plot is much lower than for indices. Perhaps there is some inconsistency in formatting between these plots?

Ex: `plot_landings(hake, group = "none")`

Image

R/save_all_plots.R
```diff
@@ -253,6 +252,7 @@ save_all_plots <- function(
         unit_label = catch_unit_label,
         scale_amount = catch_scale_amount,
         proportional = proportional,
+        interactive = TRUE,
```
sbreitbart-NOAA:
```suggestion
        interactive = interactive,
```

sbreitbart-NOAA (commented):
We're getting closer!

When testing, I found that some examples didn't produce the expected plots, nor did the plots behave as anticipated when `group = "none"`. Happy to review again once we get some more info about what might be causing those bugs. 🐛

R/plot_indices.R
```diff
@@ -1,200 +1,118 @@
...
-#       object = final,
-#       caps_alttext = caps_alttext,
-#       figures_tables_dir = figures_dir,
-#       topic_label = topic_label,
-#       fig_or_table = fig_or_table
-#     )
-#   }
-#   final
-# }
+#' Plot Index of Abundance
```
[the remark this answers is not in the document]
Schiano-NOAA: Great catch! This plot is not functioning as intended and I have figured out the issue! Will be pushing the fix shortly

R/plot_fishing_mortality.R
sbreitbart-NOAA: Ah, ok. In that case, since this code is for the `plot_fishing_mortality()` example, we should include ` module = "TIME_SERIES"` in the code so that the correct module is chosen even though `interactive = FALSE`.
And woops, I meant to write `plot_natural_mortality()`! I was referring to the fact that setting `group = "none"` still results in faceting. But if we don't expect `group = "none"'` to prevent faceting there too, then we can ignore that comment.

PR #161, opened 2025-12-03
Change "cap" to "caption" in rda (and related files)

sbreitbart-NOAA:
For the sake of consistency we are changing the name "cap" in our rda package to "caption". This is also an effort to reduce the amount of things we make as acronyms or abbreviations.

### DIFF
--- a/vignettes/manual.Rmd
+++ b/vignettes/manual.Rmd
@@ -40,7 +40,7 @@ Directly edit the rda's caption or alt_text objects. To do so, assign a new valu
 
 2. Load it into your R environment. You can do this by clicking the file or by using a command like this: `load("biomass_figure.rda")`. It will probably be called `rda`.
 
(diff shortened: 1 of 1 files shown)


Schiano-NOAA (approved):
Looks good to me! I had copilot also review it and they found an instance of cap in other documentation and added a suggestion to replace (which I accepted). Thanks for doing this so quickly!

sbreitbart-NOAA:
> Looks good to me! I had copilot also review it and they found an instance of cap in other documentation and added a suggestion to replace (which I accepted). Thanks for doing this so quickly!

Thanks, Sam! Cool that copilot saw the instance I missed. I'll merge this once [the asar PR](https://github.com/nmfs-ost/asar/pull/394#issuecomment-3609048390) is approved and merged into main

PR #162, opened 2025-12-03
Changes for FIT R universe

k-doering-NOAA:
This PR makes changes so that the primary version of stockplotr on R universe is included in the [FIT R universe](https://noaa-fisheries-integrated-toolbox.r-universe.dev/)


Schiano-NOAA (approved):
Looks good -- I think it needs a spell check and might need a word added

Issue #172, opened 2026-01-07
Change naming to stock recruitment


Schiano-NOAA:
The proper name for the stock recruitment relationship is not spawn recruitment.

This will need to be changed in all associated instance such as
- the function name
- documentation of the function
- caption and alt text if applicable

PR #174, opened 2026-01-08
Convert "spawning recruitment" to "stock recruitment" 

sbreitbart-NOAA:
Convert "spawning recruitment" to "stock recruitment" as per https://github.com/nmfs-ost/stockplotr/issues/172


Schiano-NOAA (commented):
The changes you made look good. Are there any instances where its referred to incorrectly in the caption or alt text? I requested a copilot review because it might be able to find these things.

PR #181, opened 2026-01-22
Format tables to merge error to estimates

Schiano-NOAA:
Created a new function that adds a step into creating a table that takes the processed data and
(1) formats the column names to what the final result will be
(2) Merges the error values into the estimate column to show `value (error)` and reduce the number of columns in the table

Tasks performed:
- [x] created `merge_error()` function
- [x] adjusted process_table so it exports a list of dfs for *n* number of input data frames (including 1)
- [ ] adjust create_rda to either a) create an rda for each table or b) create an rda where table is a list of tables/contains a longer list of tables within the rda
- [x] Updated `table_landings()` tests (partially failing due to issues with create_rda & the object type output being a gt? Not sure if that check is correct


sbreitbart-NOAA (approved):
This looks great! The tables look as expected for single datasets.

I think we should add a note about how this is not yet functional for lists of datasets and save that work for after NSAW. It will take some time to update `create_rda()` to work with a list of tables, since it will affect how alt text/captions are generated (each table will need the cap/alt text to identify its specific dataset). Then, `asar::create_tables_doc()` may need adjusting to work with this.

Issue #185, opened 2026-01-29
plot_biomass plots a timeseries of landings when provided the example data


Ovec8hkin:
The `plot_biomass(..., module="TIME_SERIES")` function, when applied to the example data (e.g., `example_data.rda`), appears to plot a time series of landings as opposed to a time series of stock biomass (e.g. weight of individuals in the populations, as the function name would seemingly imply). This appears to be related to the fact that "biomass_retained" and "biomass_dead" are also returned, and are what end up being plotted. 

If this is intentional, the function name and function documentation are not clear that that is intended behavior. 

Additional tests with other modules also did not yield the expected population biomass timeseries.

PR #194, opened 2026-02-23
Fix plot_biomass and update naming convention for catch in SS3 converter

Schiano-NOAA:
# Goal:

Fixing a bug identified in #185 which initially (and incorrectly) displayed biomass but it was actually catch. This makes sense since the landings and biomass plots were extremely similar.

# What has been changed:

The following fixes are implemented in this PR:

- Adjust extraction of biomass from the standard output so it only pulls biomass
- Adjust standard naming conventions in the converter to be accurate to the true quantities/estimates
- Minor addition to example in table_landings (reviewer can ignore this)

The resulting plot should show biomass over time with error as a cloud if applicable. Another issue was identified when plotting multiple model results. Feel free to test on multiple results, but the primary focus is the accuracy in the estimate plotted. Issues regarding plotting multiple output on the same plot will be addressed in a new PR.

# Actions:

To review this PR:

- Test the `convert_output()` function on a Report.sso file
- Test the `plot_biomass()` function using the example data sett found in the package and converted output in the first bullet
- Test plot_biomass(dat = list("a"=data1, "b"=data2)) to see if any errors occur when multiple datasets are provided

The maintainers @Schiano-NOAA and @sbreitbart-NOAA will handle any package checks and testing errors that might occur and make adjustments if there are any more issues found after these changes were made. Thanks for contributing the {stockplotr}!


Ovec8hkin:
@Schiano-NOAA I get the correct plots when using the Report.sso file you linked. However, the `example_data` still yields a plot of landings when `plot_biomass` is called. I would HIGHLY recommend updating the `example_data.rda` file so that the biomass plot is as expected.

sbreitbart-NOAA (approved):
Looks good!

Ovec8hkin:
@Schiano-NOAA @sbreitbart-NOAA thanks for the fix on this.

PR #195, opened 2026-02-25
Fix codebase from variety of issues

Schiano-NOAA:
# Summary

Various small fixes identified from the following issues:

* #156 
* #182 
* #183 
* #184 
* in addition, fixed a bug in `plot_spawning_biomass(...)` when facet is returned from` process_data(...)` as character(0)

PR #196, opened 2026-02-25
[Hotfix]: issue with plots when both model and grouping are present

### LINKED ISSUE #188: :bug: in `plot_timeseries()` with multiple models
@okenk noticed jagged uncertainty intervals when plotting two FIMS models that we slightly investigated.

@k-doering-NOAA originally posted this [issue on FIMS](https://github.com/NOAA-FIMS/FIMS/issues/1177). That I have since closed because I am transferring it here but it would not let me do a transfer across organizations ☹️.

>Thanks to Kiva for pointing out that the uncertainty on the last plot in the FIMS intro vignette (in this section: https://noaa-fims.github.io/FIMS/articles/fims-demo.html#length) looks a little funny - the uncertainty appears "jagged". Maybe the code generating the plot needs a fix?

screenshot of the plot having the issue

I believe the problem is in the plotting of the uncertainty where the groups are created for the lines but not for the uncertainty intervals. This could be completely off base though because I did not dig too deep. Happy to help if help if wanted.

Schiano-NOAA:
Addressing the issue brough up in #188 

An interaction between model and group was introduced in the fill and line aesthetics of the plot. We could not find a solution to renaming the legend title and decided to not do a manual adjustment to the columns to change the labels.


Schiano-NOAA:
This will be included with the release notes when #195 is merged

PR #220, opened 2026-04-03
[Fix] bugs identified in plot_recruitment

### LINKED ISSUE #211: `plot_recruitment` does not group lists
I am running
```
stockplotr::plot_recruitment(
  list(
    "asap" = output_asap,
    "fims" = output_fims
  )
)
```
and I figured that I would get two lines, one for each element of the list. Instead, I am just getting one line. The problem, I think, is in this section of `process_data()`
```
  if (is.null(group)) {
    data <- dplyr::mutate(data, group_var = "1")
  } else if (length(unique(data[[group]])) == 1) {
    data <- dplyr::mutate(data, group_var = "1")
    group <- NULL
  } else {
    if (group != "none") {
      data <- dplyr::mutate(data, group_var = .data[[group]]) |>
        dplyr::filter(!is.na(group_var))
    } else {
      # change group back to NULL bc no longer needs data processing
      group <- NULL
    }
  }
```
where group is NULL so it assigns group_var to 1 and plots everything as one group. But, I was assuming that because I put my data in as a list that I would get two lines and that I did not have to do anything additional such as assign some faceting or grouping.

### LINKED ISSUE #215: `plot_recruitment()` plots black lines for all models in list
https://github.com/nmfs-ost/stockplotr/blob/05a640552e7eb9852131dc16dff862fcfdd0932b/R/plot_recruitment.R#L101
leads to only black lines being plotted no matter how many models you supply to the `dat` input argument. Additionally, I get the following warning when I pass two models in a list to `dat`
```
Ignoring unknown labels:
* colour : "Model"
```

Schiano-NOAA:
Address the bugs identified in plot recruitment (#211 and #215)

- A leftover hard-coded value from testing was in the function. This has now been removed.
- A condition for models that properly sets grouping was added including testing conditions when indexing variables are indicated, but don't actually have unique values.
- the plot was also changed to only search for "recruitment$"

### DIFF
--- a/R/plot_recruitment.R
+++ b/R/plot_recruitment.R
@@ -104,8 +104,8 @@ plot_recruitment <- function(
     xlab = "Year",
     ylab = recruitment_label,
     group = group,
-    facet = facet #,
-    # ...
+    facet = facet,
+    ...
   ) +
     theme_noaa()
 
--- a/R/plot_recruitment.R
+++ b/R/plot_recruitment.R
@@ -35,7 +35,8 @@ plot_recruitment <- function(
   interactive = TRUE,
   module = NULL,
   make_rda = FALSE,
-  figures_dir = getwd()
+  figures_dir = getwd(),
+  ...
 ) {
   # TODO: Fix the unit label if scaling
   recruitment_label <- label_magnitude(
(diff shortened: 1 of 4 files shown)


sbreitbart-NOAA (approved):
Looks good to me!

PR #221, opened 2026-04-03
[Fix] plot_indices

### LINKED ISSUE #212: `plot_indices()` cannot accommodate a numeric group argument
Within FIMS, we have module == "Fleet" and right now we are not automatically able to pass fleet names from R to the C++ so we are relying on module_id to differentiate between the fleets. So, I passed module_id to `stockplotr::plot_indices()` as the group and I got the following error because module_id is a numeric value. It would be great if {stockplotr} could accommodate numeric groups where it just turns them into a character for you.

```
stockplotr::plot_indices(
  dat = dplyr::mutate(output_fims,
    label = dplyr::if_else(label == "index_expected", "indices", label)
  ) |>
    dplyr::filter(!is.na(observed)),
  group = "module_id"
)

Error in `ggplot2::geom_point()`:
! Problem while computing aesthetics.
i Error occurred in the 1st layer.
Caused by error in `scale_f()`:
! A continuous variable cannot be mapped to the shape aesthetic.
i Choose a different aesthetic or use `scale_shape_binned()`.
Run `rlang::last_trace()` to see where the error occurred.
```

### LINKED ISSUE #214: `plot_indices()` shows observed as line instead of points
It is pretty common to plot the observed time series as points rather than a line. I am trying to compare the observed values to two separate model fits, one from ASAP and one from FIMS. Note that I wrote my own short converter for the ASAP model. This figure is what I am seeing

Image
where it is hard to see what the observed values are and compare them to the model fits. I think it would be nice of all observed values across all of the figure were points instead of lines.

Here is what it looks like for the example data

Image

Schiano-NOAA:
Make changes per issues #212 & #214 

A new function gets created so that we can do the same plotting for other observed and predicted plot comparisons.

I also made an adjusted to process data so that the resulting df indexing variables are all character format.


Schiano-NOAA:
@sbreitbart-NOAA for some reason, exporting the rda for plot_indices is not working now. Could you ID the issue and potentially fix it? It might be due to the facetting, but I'm not sure.

Here is an example:

```
plot_indices(example_data, make_rda = TRUE)
Error in if (file.exists(fs::path(dir, "captions_alt_text.csv"))) { : 
  argument is of length zero
```

sbreitbart-NOAA:
> @sbreitbart-NOAA for some reason, exporting the rda for plot_indices is not working now. Could you ID the issue and potentially fix it? It might be due to the facetting, but I'm not sure.
> 
> Here is an example:
> 
> ```
> plot_indices(example_data, make_rda = TRUE)
> Error in if (file.exists(fs::path(dir, "captions_alt_text.csv"))) { : 
>   argument is of length zero
> ```

The plot_indices.R file isn't up-to-date with that on [the main branch](https://github.com/nmfs-ost/stockplotr/blob/main/R/plot_indices.R), which contains the new key quantities extraction/insertion pipeline. Try a rebase?

sbreitbart-NOAA (commented):
The updates look good to me but the tests are failing. Does the example dataset need to be updated? The first test passes with the `stockplotr::example_data` whereas it fails with the test's dataset.

Schiano-NOAA:
@sbreitbart-NOAA I'll check. My checks passed but something happened with my other branch where a change got overwritten so maybe this happened here or something

Schiano-NOAA:
@sbreitbart-NOAA I found the issue! The example data in the test only had one fleet which caused an error in faceting. I made a check if there as only one fleet (or grouping) to not do this action.

The checks are all passing for me now and the cmd check was good. Good to merge?

sbreitbart-NOAA:
> @sbreitbart-NOAA I found the issue! The example data in the test only had one fleet which caused an error in faceting. I made a check if there as only one fleet (or grouping) to not do this action.
> 
> The checks are all passing for me now and the cmd check was good. Good to merge?

Great! Good to merge 👍

PR #227, opened 2026-04-09
Address plot_recruitment_deviations() documentation review suggestions

### LINKED ISSUE #198: [Documentation Review] plot_recruitment_deviations
**This is an activity for the next steering committee meeting on March 10th**

# Directions

Please comment on the follow topics regarding documentation of the function [`plot_recruitment_deviations`](https://github.com/nmfs-ost/stockplotr/blob/main/R/plot_recruitment_deviations.R):

* Title
* Description
* Arguments
* Value
* Examples

You can access the documentation quickly and easily by running `?stockplotr::plot_recruitment_deviations` in your console. Remember to only focus on the documentation. If you have any feedback regarding the plot, we can/will address that in the future. Feel free to make a note of it in our meeting notes if something comes up.

Thank you for your feedback!

sbreitbart-NOAA:
Address most suggestions as per https://github.com/nmfs-ost/stockplotr/issues/198; remainders will be own issues, applicable to many other plots

@Schiano-NOAA I moved some info from 'return' to 'details'. If you're on board with this, then we should move the same information into a 'details' section for each plot

### DIFF
--- a/R/plot_recruitment_deviations.R
+++ b/R/plot_recruitment_deviations.R
@@ -1,11 +1,19 @@
-#' Plot Recruitment Deviations
+#' Plot recruitment deviations
 #'
 #' @inheritParams plot_spawning_biomass
 #'
-#' @return Plot recruitment deviations relative to one over time from an
-#' assessment model output file translated to a standardized output (\link[stockplotr]{convert_output}). There are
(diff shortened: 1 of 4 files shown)


Schiano-NOAA (approved):
💯

PR #230, opened 2026-04-09
Scale uncertainty intervals if relative = TRUE for B and SB plots

### LINKED ISSUE #218: [Bug]: plot_biomass not plotting relative error values when relative = TRUE
library(asar)

load("Western_Gulf_of_Maine_cod2024.rda")
wgom <- model_results

stockplotr::plot_biomass(
  dat = wgom,
  geom = "line",
  group = NULL,
  facet = NULL,
  #ref_line = c("target" = 10),
  unit_label = "mt",
  scale_amount = 1,
  relative = TRUE, # changed from FALSE
  #interactive = FALSE,
  #module = "TIME_SERIES",
  linewidth = 2 # increase width of lines
) +
  ggplot2::geom_vline(xintercept = 2005, linetype = "dashed", color = "red")

sbreitbart-NOAA:
Address https://github.com/nmfs-ost/stockplotr/issues/218


R/plot_biomass.R
```diff
@@ -121,7 +121,10 @@ plot_biomass <- function(
     }
     if (is.na(ref_line_val)) cli::cli_abort("Reference value not found. Cannot plot relative values.")
     prepared_data <- prepared_data |>
-      dplyr::mutate(estimate = estimate / ref_line_val)
+      dplyr::mutate(estimate = estimate / ref_line_val,
+        estimate_lower = estimate_lower / ref_line_val,
+        estimate_upper = estimate_upper / ref_line_val
```
kellijohnson-NOAA: @sbreitbart-NOAA I am not sure if I fully understand what is going on here because I have not looked too terribly hard at the code but I just wanted to let you know that the uncertainty should be the uncertainty based on the actual derived quantity and not just the scaled uncertainty. For example, the uncertainty of biomass comes from the calculation of the confidence intervals of biomass using the standard deviation of biomass estimate. The same should be true for unfished biomass, where the uncertainty of what is being plotted should be from a standard deviation.

Schiano-NOAA: I am not following. Here we are scaling down the uncertainty when relative = TRUE where all values on the plot are getting scaled by the reference line value. Are you saying that here the confidence intervals should be based on the error for the reference line? 

Are the current calculation for error correct in filter_data?
https://github.com/nmfs-ost/stockplotr/blob/0397a61a4baefb7fe05c7b429553457dc742fe43/R/utils_plot.R#L680-L689

After review, I think they aren't correct except for SE, but I get confused with error calcs.

sbreitbart-NOAA: @kellijohnson-NOAA thank you for noticing this- I remember discussing this a while back. I don't know the proper method to use here. Maybe the steering committee would know?

kellijohnson-NOAA: If you were multiplying or dividing by a constant it would be fine but we aren't instead we are dividing my a calculated value that has its own uncertainty. There are ways to calculate the uncertainty of the result but more typically, you just output the derived quantity, e.g., time series of spawning_biomass/unfished_biomass, as another derived quantity and you get the uncertainty of the time series.

Schiano-NOAA: Ah okay I see. This is jogging my memory now thanks!

Schiano-NOAA (changes requested):
Putting this on halt until we can talk to the steering committee and also incorporate the calcs for error here

Schiano-NOAA:
@sbreitbart-NOAA can you review this? 😄

R/plot_biomass.R
```diff
@@ -74,55 +75,48 @@ plot_biomass <- function(
     scale_amount <- 1
   }
 
-  # Filter data for spawning biomass
+  # Filter data for biomass
+  # TODO: determine method to ID that first point in the timeseries is actually Bunfished ref pt
   prepared_data <- filter_data(
     dat = dat,
-    label_name = "^biomass$",
+    label_name = ifelse(relative, "biomass_biomass_unfished|biomass_ratio", "^biomass$"), # what exactly is biomass_ratio?
```
sbreitbart-NOAA, line 82: Did you find out what biomass_ratio is? 😄

R/plot_fishing_mortality.R
```diff
@@ -89,39 +78,18 @@ plot_fishing_mortality <- function(
...
-    } else {
-      F.min <- min(prepared_data$estimate) |> round(digits = 3)
-      F.max <- max(prepared_data$estimate) |> round(digits = 3)
-
-      export_kqs(F.min, F.max)
-      insert_kqs(F.min, F.max)
-    }
-
+    
+    F.min <- min(prepared_data$estimate) |> round(digits = 3)
```
sbreitbart-NOAA, line 87: Did you remove the calculations for rel.F.min and rel.F.max? These need to be calculated for insertion into the caps_alt_text.csv and key_quantities.csv

R/plot_fishing_mortality.R
```diff
@@ -140,7 +108,7 @@ plot_fishing_mortality <- function(
 
     create_rda(
       object = final,
-      topic_label = ifelse(relative, "relative_fishing_mortality", "fishing_mortality"),
+      topic_label = "fishing_mortality", # ifelse(relative, "relative_fishing_mortality", "fishing_mortality"),
```
sbreitbart-NOAA, line 111: the topic_label should be relative_fishing_mortality if relative = TRUE to match up with the label in captions_alt_text.csv

R/plot_fishing_mortality.R
Schiano-NOAA: Yes because I removed it from the function

R/plot_fishing_mortality.R
Schiano-NOAA: I took out relative as an argument so this was removed. Is there any other adjustment that needs to be made for the kqs?

sbreitbart-NOAA: Oh, I see now! Yes- the captions_alt_text_template.csv and key_quantity_template.csv (in the inst folder) should be updated to remove the rows associated with relative F mortality and its relative key quantities

sbreitbart-NOAA:
Looks good to me! Thanks for updating those csvs 👍

PR #236, opened 2026-04-24
[Hotfix] change the default era to NULL in multiple functions

Schiano-NOAA:
Setting default era to NULL instead of  "time" so nothing is unnecessarily filtered by default

Plan release for v0.11.1


R/plot_stock_recruitment.R
```diff
@@ -27,7 +27,7 @@ plot_stock_recruitment <- function(
   spawning_biomass_label = "mt",
   recruitment_label = "mt",
   interactive = TRUE,
-  # era = "time",
+  era = "time",
```
sbreitbart-NOAA: Should this be NULL, just like for the other plotting functions?

Schiano-NOAA: No I had set it NULL but it made the tests fail so something is wack when it's set to NULL

sbreitbart-NOAA: Do you know what's unexpectedly happening? It's strange that everything else works when this argument is set consistently, except for this one

Schiano-NOAA: I think in this case, it needs era = "time" because it has that in the values? But I would have to look into it to be sure

Schiano-NOAA: I found this issue. I opened a new PR which needs one additional thing but this turned out to be more complex (per usual). However, I think this is good once the fix is made

sbreitbart-NOAA: Great! I'm sure that addressing this issue will increase the consistency and robustness of our plotting pipeline's architecture.

Schiano-NOAA: I requested review to address this in another PR which will be merged into this one. Then it can get merged into main. I am going to merge the dev branch then rebase this branch

Edit: actually I think we should merge this into dev then push out a release once the 3 open working PRs are approved  (which I think should be relatively good)

PR #237, opened 2026-04-27
[Feature]: add option to convert plot units to lbs

### LINKED ISSUE #223: Add options for plots to be output in lbs
For regions who's stakeholders prefer to see the plots' axis in lbs, it would be good to add an argument to do this so it can scale for lbs.

**Need to check if this is only for landings/catch only**

Schiano-NOAA:
New argument in spawning_biomass, biomass, and landings plot function to convert units from metric tons to lbs

Plan release for v0.11.0


Schiano-NOAA:
@sbreitbart-NOAA this is ready for review and you will receive a request shortly. I wanted to note first that

- bam exports landings in klbs (thousands of lbs) so this will be on the user to change unit_label and NOT use the lbs argument
- error will not be present when converting from kgs to lbs
- this argument assumes the output is in kgs

We should plan for this to be temporary until the units column is incorporated into the converter for all models. Once that feature is developed, we can adjust the lbs feature in respective plots to pull the units from the conout file and proceed from there (i.e. if units are already lbs and user sets lbs = TRUE then nothing happens and unit_label can be extracted from the model)

sbreitbart-NOAA (commented):
Github won't let me comment on individual lines, so I'll add my comments here. Looks like this is working well! Just have a couple of suggestions and a question:

plot_spawning_biomass.R
L39: suggest updating param to something like this:
`#' @param lbs A logical value indicating whether to convert the y-axis values, assumed to be in kilograms, to pounds.`
L135: Did you end up verifying this, as per your comment?

process_data.R
L13: Same comment as for plot_spawning_biomass.R L39

Schiano-NOAA:
@sbreitbart-NOAA Maybe I should changed assumed to default. We are just setting the default to FALSE saying the output units are in lbs.

sbreitbart-NOAA:
> @sbreitbart-NOAA Maybe I should changed assumed to default. We are just setting the default to FALSE saying the output units are in lbs.

I'm not sure I understand. Could you please rephrase?

Schiano-NOAA:
> > @sbreitbart-NOAA Maybe I should changed assumed to default. We are just setting the default to FALSE saying the output units are in lbs.
> 
> I'm not sure I understand. Could you please rephrase?

My apologies. I misunderstood what you said in your first comment. I can add some indication that we assume the default units are kgs

Schiano-NOAA:
@sbreitbart-NOAA okay thoughts on the new wording?

sbreitbart-NOAA:
> @sbreitbart-NOAA okay thoughts on the new wording?

So close! I think you're just missing a word ("from"):

@param lbs A logical value indicating whether to convert the y-axis values **from**
#' kilograms to pounds. The default units match the default in the
#' unit_label argument - 'metric tons'.

PR #243, opened 2026-05-05
[Revamp] sr plot use process_data()

Schiano-NOAA:
The stock recruitment plot does not currently use process data and is not working when certain conditions aren't met. This PR make the function more flexible and revamped it on the back end.

Remaining to do:
- [x] add condition for when expected/predicted recruitment are present and line is plotted on top
- [x] test

### DIFF
--- a/man/plot_stock_recruitment.Rd
+++ b/man/plot_stock_recruitment.Rd
@@ -9,7 +9,7 @@ plot_stock_recruitment(
   spawning_biomass_label = "mt",
   recruitment_label = "mt",
   interactive = TRUE,
-  era = "time",
+  era = NULL,
   module = NULL,
   scale_amount = 1,
   make_rda = FALSE,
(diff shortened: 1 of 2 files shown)


sbreitbart-NOAA (commented):
Github still won't let me comment on individual lines, so I'll add my comments here. Looks like this is working well! Just one suggestion:

plot_biomass.R
L59: Default should be changed to NULL for this and other functions

Schiano-NOAA:
@sbreitbart-NOAA I am not sure what line your referring to because I rebased. I will need to make this change for the other branch though because this branch does not have changes in plot_biomass

man/plot_stock_recruitment.Rd
sbreitbart-NOAA, line 37: This should be NULL instead of "time", based on the function's default arguments

Schiano-NOAA: This gets updated in the other PR

PR #246, opened 2026-05-06
Update documentation: move info from 'returns' to 'details'

### LINKED ISSUE #244: Move info from return into details for all plots, following convention in https://github.com/nmfs-ost/stockplotr/pull/227

sbreitbart-NOAA:
Address https://github.com/nmfs-ost/stockplotr/issues/244; replace 'return' with 'returns'

### DIFF
--- a/R/add_theme.R
+++ b/R/add_theme.R
@@ -4,7 +4,8 @@
 #' @param discrete TRUE/FALSE indicate if data input is discrete. Specifically applied to plots.
 #'
 #' @returns Add the standard formatting for stock assessment reports for any
-#' figure or table. Currently, the function is able to format objects from:
+#' figure or table.
+#' @details Currently, the function is able to format objects from:
 #' ggplot (ggplot2), base r plot, flextable (flextable), gt tables (gt), and kable tables (kableExtra).
 #' @export
 #'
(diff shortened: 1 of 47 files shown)


Schiano-NOAA (changes requested):
Looks good for the most part! Why are we using returns instead of return?  and this needs a rebase to dev. There are parts in here from other PRs along with quarto getting added back in as a dependency.

(maybe dev needs a rebase, but I might have just done that in our office hour)

sbreitbart-NOAA:
> Looks good for the most part! Why are we using returns instead of return? and this needs a rebase to dev. There are parts in here from other PRs along with quarto getting added back in as a dependency.
> 
> (maybe dev needs a rebase, but I might have just done that in our office hour)

Ah, I think it just needed some more updating and rebasing 👍 
We had been using a mix of "return" and "returns". The former was superseded by the latter, so I thought it'd be helpful to use one term instead of both.

Schiano-NOAA:
> > Looks good for the most part! Why are we using returns instead of return? and this needs a rebase to dev. There are parts in here from other PRs along with quarto getting added back in as a dependency.
> > (maybe dev needs a rebase, but I might have just done that in our office hour)
> 
> Ah, I think it just needed some more updating and rebasing 👍 We had been using a mix of "return" and "returns". The former was superseded by the latter, so I thought it'd be helpful to use one term instead of both.

Sounds good! I wonder if the automatic doxygen insertion has this update as well

Schiano-NOAA (approved):

PR #251, opened 2026-05-08
[Fix] process dat issue with facetting

Schiano-NOAA:
This fixes an issue I identified when facetting is explicit in the plot_indices function. 

I made adjustments to automatic detection and assignment of facet in the process_data function specifically so it does not overwrite facet or include the facet as group when it's the first indexing found in the index_variable object

plan release for v0.11.1


sbreitbart-NOAA (approved):
Looks good!

PR #263, opened 2026-05-19
[feat] add options to scale and round table values

### LINKED ISSUE #254: [Feat] Add argument to round values to x decimal
Combine with PR to add new argument for scaling values in tables (same release)

### LINKED ISSUE #255: [Feat] add argument to scale values in tables
Combine with PR to add new argument for decimal values in tables (same release)

Schiano-NOAA:
Addresses suggestions from workshops and issues #254 & #255

### DIFF
--- a/R/table_landings.R
+++ b/R/table_landings.R
@@ -16,6 +16,10 @@
 #' Default: "sum"
 #'
 #' Options: "sum" or "mean"
+#' @param digits Numeric value indicating the number of digits values in the 
+#' table will be rounded to.
+#' 
+#' Default: 2
 #' @param tables_dir The location of the folder containing the generated table
 #' rda files ("tables") that will be created if the argument `make_rda` = TRUE.
 #' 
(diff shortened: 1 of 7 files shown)


R/table_landings.R
```diff
@@ -6,6 +6,9 @@
 #' want to just summarize the data across all factors, set group = "none".
 #' @param method A string describing the method of summarizing data when group
 #' is set to "none". Options are "sum" or "mean". Default is "sum".
+#' @param digits Numeric value indicating the number of rounding digits.
```
sbreitbart-NOAA:
```suggestion
#' @param digits Numeric value indicating the number of rounding digits. Trailing zeroes will not be added to the estimate (example: rounding 0.1 to 4 digits will still produce 0.1).
```

This doesn't add trailing zeroes if the number doesn't have them (example: `round(as.numeric(5.5), digits = 4)` will still produce 5.5. I suggest noting that here

sbreitbart-NOAA (commented):
I had one minor suggestion. Also, did you mean to reference this issue in the description? (254 is repeated): https://github.com/nmfs-ost/stockplotr/issues/255

Otherwise, looks good! Minor changes but substantial benefits for users!

Schiano-NOAA:
> I had one minor suggestion. Also, did you mean to reference this issue in the description? (254 is repeated): #255
> 
> Otherwise, looks good! Minor changes but substantial benefits for users!

LOL yes it was supposed to be 255 as well. I will make those changes then merge! Thanks for the review 😄

R/table_landings.R
Schiano-NOAA: I wonder if there is a arg in round to still show trailing zeros. I am going to look into it before I merge this suggestion.

Schiano-NOAA: I have fixed this!

PR #264, opened 2026-05-20
[docs] clarify documentation for abundance at age plot

### LINKED ISSUE #199: [Documentation Review] plot_abundance_at_age
**This is an activity for the next steering committee meeting on March 10th**

# Directions

Please comment on the follow topics regarding documentation of the function [`plot_abundance_at_age`](https://github.com/nmfs-ost/stockplotr/blob/main/R/plot_abundance_at_age.R):

* Title
* Description
* Arguments
* Value
* Examples

You can access the documentation quickly and easily by running `?stockplotr::plot_abundance_at_age` in your console. Remember to only focus on the documentation. If you have any feedback regarding the plot, we can/will address that in the future. Feel free to make a note of it in our meeting notes if something comes up.

Thank you for your feedback!

### LINKED ISSUE #259: plot_abundance_at_age facet = NULL not working
When I run `plot_abundance_at_age()` and set `facet = NULL` there are still facets, but if I set `facet = 'none'` it does work. I would have expected `facet = NULL` to provide the same result as `'none'`

Schiano-NOAA:
adjust naa documentation and respective inherited docs in response to feedback on #199 & #259

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -71,3 +71,4 @@ Encoding: UTF-8
 Language: en-US
 LazyData: true
 Config/roxygen2/version: 8.0.0
+RoxygenNote: 7.3.3
--- a/R/plot_abundance_at_age.R
+++ b/R/plot_abundance_at_age.R
@@ -1,4 +1,4 @@
-#' Plot Abundance (or Numbers) at Age (AAA or NAA)
+#' Plot Abundance at Age
 #'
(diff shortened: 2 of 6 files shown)


sbreitbart-NOAA:
I suggest rebasing this branch to the add-defaults branch in this PR https://github.com/nmfs-ost/stockplotr/pull/267. It also updates documentation, but for all functions. Then, we could ensure consistency in that, e.g., changes in this PR would also be applied to the other functions, and vice versa.

PR #267, opened 2026-05-20
[doc] Add default and optional values for arguments in documentation

### LINKED ISSUE #256: [Docs] Add defaults for each argument in Roxygen as in plot_spawning_biomass()

sbreitbart-NOAA:
Add default and optional values for arguments in documentation as per https://github.com/nmfs-ost/stockplotr/issues/256

### DIFF
--- a/R/save_all_plots.R
+++ b/R/save_all_plots.R
@@ -8,7 +8,7 @@
 #' recruitment_scale_amount = 100 would scale down a value from 500,000 -->
 #' 5,000. This scale will be reflected in the y axis label.
 #' 
-#' Default: "1"
+#' Default: 1
 #' 
 #' @param recruitment_unit_label Units for recruitment
 #' 
@@ -36,7 +36,7 @@
 #' @param biomass_scale_amount A number describing how much to scale down the
 #' biomass quantities shown on the y axis. See `recruitment_scale_amount`.
 #' 
-#' Default: "1"
+#' Default: 1
 #' 
 #' @param landings_unit_label Units for landings
 #' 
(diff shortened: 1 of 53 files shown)


R/save_all_plots.R
```diff
@@ -7,9 +7,18 @@
 #' the recruitment quantities shown on the y axis. For example,
 #' recruitment_scale_amount = 100 would scale down a value from 500,000 -->
 #' 5,000. This scale will be reflected in the y axis label.
+#' 
+#' Default: "1"
```
Schiano-NOAA: The default is not a character

R/save_all_plots.R
```diff
@@ -19,28 +28,82 @@
...
 #' indicate the reference line in the form c("label" = value).
+#' 
+#' Default: "msy"
+#' 
+#' Options: Including, but not limited to: "target", "MSY", "unfished"
+#' 
 #' @param biomass_scale_amount A number describing how much to scale down the
 #' biomass quantities shown on the y axis. See `recruitment_scale_amount`.
+#' 
+#' Default: "1"
```
Schiano-NOAA: ditto to above and rest of scale_amounts

Schiano-NOAA (changes requested):
Thanks for doing this! It generally looks good. Only change that needs to be made is that scale_amount for I believe all functions is a numeric value and the default should be 1 not "1"

Also in regards to your comment in #264, instead of rebasing that branch to this one,  I think we should just mege this one when it's finished then I can rebase to main on the other branch

PR #272, opened 2026-05-29
[fix] convert_output() with updated SS3 report file

Schiano-NOAA:
This PR fixes the issues users were encountering when convert_output() was failing with the most recent version of SS3.

This should be now fixed and still work as previously expected.


Schiano-NOAA:
@iantaylor-NOAA could test out convert_output() on this branch with the newest versions of SS3? I tested it out with an example Jason sent me and it works not, but doesn't hurt to test on other stocks!

sbreitbart-NOAA (approved):

PR #274, opened 2026-06-02
[feat] Update indices table

### LINKED ISSUE #273: [Revamp]: `table_index()`

sbreitbart-NOAA:
As per https://github.com/nmfs-ost/stockplotr/issues/273

@Schiano-NOAA In my testing, I see that the estimates can get pretty large. I think it'd be useful to format large numbers with commas, like with the [`format` function](https://rguides.dev/cookbooks/how-to-format-numbers-with-commas/). I'm not sure where to put this though; if I try in [line 76](https://github.com/nmfs-ost/stockplotr/pull/274/changes#diff-7306ef2829c592ff715ed7e8b416c88343742c69fc1e1cd704025f65e17a51e2R76), it causes issues downstream. Thoughts?


Schiano-NOAA:
> As per #273
> 
> @Schiano-NOAA In my testing, I see that the estimates can get pretty large. I think it'd be useful to format large numbers with commas, like with the [`format` function](https://rguides.dev/cookbooks/how-to-format-numbers-with-commas/). I'm not sure where to put this though; if I try in [line 76](https://github.com/nmfs-ost/stockplotr/pull/274/changes#diff-7306ef2829c592ff715ed7e8b416c88343742c69fc1e1cd704025f65e17a51e2R76), it causes issues downstream. Thoughts?

I can't remember where it is, but there is a point in the workflow (I want to say during process_table) where I had to change values from numeric to character in order to keep the trailing zeros. I would suggest doing it right after that step since I think what you are implying, the commas are being placed on a character?

R/table_indices.R
```diff
@@ -1,173 +1,186 @@
...
+        system.file("resources", "captions_alt_text_template.csv", package = "stockplotr")
+      )
+    # export df with captions and alt text to csv
+    utils::write.csv(
+      x = caps_alttext,
+      file = fs::path(getwd(), "captions_alt_text.csv"),
+      row.names = FALSE
+      )    
+    }
+
```
Schiano-NOAA: Small thing, but can you remove all of these spaces

sbreitbart-NOAA:
> > As per #273
> > @Schiano-NOAA In my testing, I see that the estimates can get pretty large. I think it'd be useful to format large numbers with commas, like with the [`format` function](https://rguides.dev/cookbooks/how-to-format-numbers-with-commas/). I'm not sure where to put this though; if I try in [line 76](https://github.com/nmfs-ost/stockplotr/pull/274/changes#diff-7306ef2829c592ff715ed7e8b416c88343742c69fc1e1cd704025f65e17a51e2R76), it causes issues downstream. Thoughts?
> 
> I can't remember where it is, but there is a point in the workflow (I want to say during process_table) where I had to change values from numeric to character in order to keep the trailing zeros. I would suggest doing it right after that step since I think what you are implying, the commas are being placed on a character?

Yes! Thanks- found it in `process_table()`. If that change looks good to you, I'll merge.

Schiano-NOAA:
@sbreitbart-NOAA this isn't working with BAM models, so I am checking on a fix for it in a sec. Did you test it for Rceattle as well? Also, could you change the name from plot_indices to plot_index? I need to make this change in the naming conventions in the converter as well and will also be for the label.

sbreitbart-NOAA:
@Schiano-NOAA This isn't working for Rceattle. I'll work on fixing the function name first

Schiano-NOAA:
@sbreitbart-NOAA plot_indices was working for Rceattle when I was working on the converter (if that helps). I used that as a comparison to the Rceattle::plot_index function to make sure values were correctly converted and the plots looked the same

Schiano-NOAA:
@sbreitbart-NOAA do you want me to change the naming in the excel files for the converter? I remember you saying you don't have excel locally

sbreitbart-NOAA:
> @sbreitbart-NOAA do you want me to change the naming in the excel files for the converter? I remember you saying you don't have excel locally

I can change them with find/replace. Wasn't sure if those should be updated but now that I know, I can change the other instances of "indices" downstream 👍

Schiano-NOAA:
@sbreitbart-NOAA great thank you! I also just put out a release (sorry for the multiple emails, I messed up the first ~couple ~ times lol). Could you also update the version in this PR to .9000 since this will probably be the first one for the next release?

Schiano-NOAA:
@sbreitbart-NOAA this should work now. Please test!

Could you also test table_landings for me too? I tried it and get this error for any test data set 

```
Error in if (is.na(uncert_lab)) uncert_lab <- "uncertainty" : 
  argument is of length zero
```

however, when I run it line-by-line, it works fine.

sbreitbart-NOAA:
A few notes:
1. The PR's checks are failing because the `table_index()` example isn't working. I updated the stockplotr example_data and it's working fine for me locally now
2. `table_index()` works with SS3, BAM, and Rceattle, but the column headers aren't exactly right for Rceattle (see the remaining underscores):
image

3. When I run `table_landings()` line by line, I still get an error at line 107. I think it's because `uncert_lab` isn't converting into an NA, but instead "named character(0)". If I change it to NA, I still get another error at line 116: 
`Error in names(object) <- nm : attempt to set an attribute on NULL`

sbreitbart-NOAA:
> A few notes:
> 
>     1. The PR's checks are failing because the `table_index()` example isn't working. I updated the stockplotr example_data and it's working fine for me locally now
> 
>     2. `table_index()` works with SS3, BAM, and Rceattle, but the column headers aren't exactly right for Rceattle (see the remaining underscores):
> 
> 
> image
> 
>     3. When I run `table_landings()` line by line, I still get an error at line 107. I think it's because `uncert_lab` isn't converting into an NA, but instead "named character(0)". If I change it to NA, I still get another error at line 116:
>        `Error in names(object) <- nm : attempt to set an attribute on NULL`

For #1, meant to say `plot_index()`, not `table_index()`!

Schiano-NOAA:
Note: as of right now, both tables are working for SS3 and BAM models. I have not yet been able to work out how to do this process for additional grouping which is causing the error for Rceattle examples. The errors are resulting from the `merge_error()` function when aligning column names and then combining error columns with their estimate counterpart.

I started on a fix, but it needs to be fleshed out and tested. The fix includes the following:

https://github.com/nmfs-ost/stockplotr/blob/b4640f3c203bf7023cdd42244bf14e71ac56c8b6/R/utils_table.R#L223-L243

https://github.com/nmfs-ost/stockplotr/blob/b4640f3c203bf7023cdd42244bf14e71ac56c8b6/R/utils_table.R#L305-L320

Schiano-NOAA:
@sbreitbart-NOAA I believe this is ready. There is an issue with Rceattle with table_landings, but this is a separate issue.

PR #277, opened 2026-06-09
[tests] Remove converted model results files and update tests to use example data

sbreitbart-NOAA:
* Remove converted model results files
* update tests to use example data


Schiano-NOAA (approved):
LGTM!

PR #278, opened 2026-06-10
[fix] Fix method for extracting plot/table name 

sbreitbart-NOAA:
Fix method for extracting plot/table name when exporting rda to allow for calling library when making plot


Schiano-NOAA (approved):
LGTM

PR #281, opened 2026-06-15
[fix]: rescaling error in plots

### LINKED ISSUE #257: uncertainty in spawning biomass plot not scaling correctly
I tried to use the `rescale` option with the example plot from the workshop. The axis and axis label changed exactly as expected, but the uncertainty interval got messed up as shown in the plot below.
```
plot_spawning_biomass(
  dat = paka, # dataset
  unit_label = "mt", # unit label: metric tons
  scale_amount = 1000, # scale to 1000s of metric tons
  relative = FALSE, # show spawning biomass, NOT relative spawning biomass
  interactive = FALSE, # prompt user for MODULE_NAME in console
  module = "DERIVED_QUANTITIES" # SS3 module which includes uncertainty
)
```
Image

Schiano-NOAA:
fix issue with error scaling in plots as identified by #257 

The issue was that the estimate was scaled before the error thus in the confidence interval calcs, the estimate value was messed up for the calcs. I moved the estimate to get scaled after the CIs were calculated and this fixed the issue.

I also removed a ton of commented out code that we no longer use in that function.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: stockplotr
 Title: Tables and Figures for Stock Assessments
-Version: 0.12.0
+Version: 0.12.1.9000
 Authors@R: c(
     person("Samantha", "Schiano", , "samantha.schiano@noaa.gov", role = c("aut", "cre"),
            comment = c(ORCID = "0009-0003-3744-6428")),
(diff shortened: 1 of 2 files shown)


sbreitbart-NOAA (approved):
Nice!

Issue #284, opened 2026-06-26
bug: No applicable method for `@` applied to an object of class `ggplot_built`


kellijohnson-NOAA:
https://github.com/nmfs-ost/stockplotr/blob/64c387e317c4345ca10a60b6ebbcf9d6a248ac22/R/plot_spawning_biomass.R#L216
is causing the following error on linux using version 0.12.1.9000 of stockplotr.

```
ggplot2::ggplot_build(plot)@data[[2]][["x"]]
Error in ggplot2::ggplot_build(plot)@data : 
  no applicable method for `@` applied to an object of class "ggplot_built"
```
but I also see the use of `@` in three files
<img width="880" height="583" alt="Image" src="https://github.com/user-attachments/assets/e1d70af6-a386-4716-ac5f-69c339e5c3f8" />

If I use a dollar sign the error goes away, `ggplot2::ggplot_build(plot)$data[[2]][["x"]]` or if I use square brackets `ggplot2::ggplot_build(plot)[["data"]][[2]][["x"]]`.

Schiano-NOAA:
Potential suggestion is use an alternative package like some kind of gg extension that can either extract the data or use the underlying data instead of using ggplot_build

Schiano-NOAA:
This was resolved in #286

PR #286, opened 2026-06-30
[fix] Refactor code to avoid Linux building issue

sbreitbart-NOAA:
Refactor code in 3 functions as per https://github.com/nmfs-ost/stockplotr/issues/284


Schiano-NOAA (approved):
LGTM! Was this only in these functions? I'm surprised we don't have it in others

sbreitbart-NOAA:
> LGTM! Was this only in these functions? I'm surprised we don't have it in others

Yes, only in these functions. Thanks!

PR #287, opened 2026-06-30
[feat] Exclude uncertainty from landings table

### LINKED ISSUE #258: Suppress SEs in `table_landings()`
While models require a small amount of flexibility in landings time series so we have non-zero standard errors, these are generally not meaningful and not usually reported. An option to include or suppress SEs would be nice.

sbreitbart-NOAA:
Adjust table_landings() workflow to exclude uncertainty as per https://github.com/nmfs-ost/stockplotr/issues/258; update test

### DIFF
--- a/R/table_landings.R
+++ b/R/table_landings.R
@@ -120,7 +120,10 @@ table_landings <- function(
     label = label,
     digits = digits
   )
-  table_data <- table_data_info[[1]]
+  table_data <- table_data_info[[1]] |>
+    # remove uncertainty columns
+    purrr:::map(~ dplyr::select(.x, !contains(uncert_lab))) |>
+    purrr:::map(~ dplyr::select(.x, !contains("uncertainty", ignore.case = TRUE)))
   indexed_vars <- table_data_info[[2]]
   id_col_vals <- table_data_info[[3]]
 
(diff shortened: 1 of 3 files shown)


kellijohnson-NOAA:
I think not having uncertainty for landings is a very region specific thing. And, I notice that this PR removes uncertainty entirely. @lisaailloud-NOAA isn't uncertainty in landings history a big part of SE assessments? I can move this conversation to the original issue if you would like @sbreitbart-NOAA. But, I think just because the NWFSC doesn't report uncertainty of landings because the assumed value is so small doesn't mean that other regions have small values. I do not see the harm of reporting small values for the NWFSC when other centers are going to want the value.

sbreitbart-NOAA:
> I think not having uncertainty for landings is a very region specific thing. And, I notice that this PR removes uncertainty entirely. @lisaailloud-NOAA isn't uncertainty in landings history a big part of SE assessments? I can move this conversation to the original issue if you would like @sbreitbart-NOAA. But, I think just because the NWFSC doesn't report uncertainty of landings because the assumed value is so small doesn't mean that other regions have small values. I do not see the harm of reporting small values for the NWFSC when other centers are going to want the value.

@kellijohnson-NOAA These are important points. I got my signals crossed and developed this feature though we (Sam and I) had previously agreed to keep it in. *Maybe* we'll make an option to remove it in the future. Sorry about this! Going to close the PR and keep the branch.

PR #301, opened 2026-07-16
[feat] Add more key quantities needed for asar SIS exporter function

sbreitbart-NOAA:
Add capability to calculate and store key quantities necessary for the in-development `asar::export_to_sis()` function


inst/resources/key_quantity_template.csv
```diff
@@ -5,9 +5,12 @@ B.max,,maximum biomass,,,,
 B.min,,minimum biomass,,,,
 B.ref.pt,,biomass limit reference point,,,,
 B.start.year,,start year of the biomass time series,,,,
+B.terminal.year,,terminal year of biomass measurement,,,,
 B.units,,units of biomass measurement,,,,
 Bend,,biomass at the end of the time series,,,,
-Bmsy,,biomass at maximum sustainable yield,,,,
+B.msy,,biomass at maximum sustainable yield,,,,
+B.msy.max,,upper estimate of the biomass at maximum sustainable yield confidence interval,,,,
```
Schiano-NOAA: maybe include of the 95th percentile CI (same with the lower)

inst/resources/key_quantity_template.csv
```diff
@@ -22,10 +25,13 @@ cpue.start.year,,start year of the CPUE time series,,,,
 cpue.units,,units for catch per unit effort,,,,
 F.FMSY.end.yr,,fishing mortality relative to FMSY in the final year,,,,
 F.end.year,,final year of the fishing mortality time series,,,,
+F.limit,,fishing mortality limit,,,,
 F.max,,maximum fishing mortality,,,,
 F.min,,minimum fishing mortality,,,,
 F.ref.pt,,fishing mortality limit reference point,,,,
 F.start.year,,start year of the fishing mortality time series,,,,
+F.target,,target fishing mortality reference point,,,,
+F.terminal.year,,terminal year of fishing mortality measurement,,,,
```
Schiano-NOAA:
```suggestion
F.terminal.year,,terminal year of fishing mortality,,,,
```

inst/resources/key_quantity_template.csv
```diff
@@ -5,9 +5,12 @@ B.max,,maximum biomass,,,,
 B.min,,minimum biomass,,,,
 B.ref.pt,,biomass limit reference point,,,,
 B.start.year,,start year of the biomass time series,,,,
+B.terminal.year,,terminal year of biomass measurement,,,,
```
Schiano-NOAA:
```suggestion
B.terminal.year,,terminal year of biomass,,,,
```

Schiano-NOAA (changes requested):
Is there any way we can take this out of the function? I think this is making the function much longer than needed and it's not relevant to making the plot since it serves a different purpose.

sbreitbart-NOAA:
> Is there any way we can take this out of the function? I think this is making the function much longer than needed and it's not relevant to making the plot since it serves a different purpose.

What code/function are you talking about, specifically? All of it?

Schiano-NOAA:
@sbreitbart-NOAA any of the key quantity calcs. I think it would be worth it to at least put them in their own function and call them in the plot_x function with the dat as input

sbreitbart-NOAA:
> @sbreitbart-NOAA any of the key quantity calcs. I think it would be worth it to at least put them in their own function and call them in the plot_x function with the dat as input

Ok, just made a new function `calc_kqs()` that calculates the key quantities in a few functions- primarily those that require more than a line to calculate- so the plotting functions aren't as long anymore. Is that sufficient re: your idea?

Schiano-NOAA:
> > @sbreitbart-NOAA any of the key quantity calcs. I think it would be worth it to at least put them in their own function and call them in the plot_x function with the dat as input
> 
> Ok, just made a new function `calc_kqs()` that calculates the key quantities in a few functions- primarily those that require more than a line to calculate- so the plotting functions aren't as long anymore. Is that sufficient re: your idea?

Yes this is great. Cleans up the code well and we can still get the quantities we need efficiently 😄

Schiano-NOAA (approved):

PR #315, opened 2026-07-28
[feat]: enable user to add a reference line to the plot for each input dat

### LINKED ISSUE #268: [fix] reference_line function throwing s4 class error for some users
There is an issue using ggplot_build function in the reference_line function where for a user it cause the function to fail and the following error:

```
Error in vapply(x, is.null, logical(1)) : 
  trying to get slot "data" from an object (class "ggplot2::ggplot_built") that is not an S4 object
```

### LINKED ISSUE #289: For all applicable plots: add ability to show ref_line for each model

Schiano-NOAA:
I moved the reference line portion of the plot into it's own function -- `add_reference_line()` -- which gets added to the plot using the (+) operator. 

This feature automatically applies to each input dat. There are 4 conditions that this feature handles 

1. reference lines and values extracted for each input dat. e.g. 
`plot_spawning_biomass(dat=list("mod1"=mod1, "mod2"=mod2), ref_line = "msy")` SBmsy is found for each model.
2. custom values are input by the user for each model in order they are listed in dat. e.g.  `plot_spawning_biomass(dat=list("mod1"=mod1, "mod2"=mod2), ref_line = c("msy"=1000, "msy"=1200))` In this instance, any name or combinations can be used as long as there is one value for each element in the dat list.
3. different reference lines for each model extracted automatically. e.g.  `plot_spawning_biomass(dat=list("mod1"=mod1, "mod2"=mod2), ref_line = c("msy","target"))`
4. Don't include lines at all. Set `ref_line = NULL`

### DIFF
--- a/R/plot_biomass.R
+++ b/R/plot_biomass.R
@@ -141,37 +141,56 @@ plot_biomass <- function(
     ...
   )
   # Add reference line
+  # Conditions for ref line
+  # 1. all are comparing same value = msy, target, unfished...
+  # 2. custom input vector = c("sable23_msy"=30, "sable25_msy"=40) -- user must indicate which model in label otherwise it will assign in order of dat
+  # 3. input vector of labels = c("msy", "target")
   # getting data set - an ifelse statement in the fxn wasn't working
   if (relative) {
     # don't add any reference line here and just add theme for final plot
     final <- plt + theme_noaa()
   } else {
-    if ("unfished" %in% c(names(ref_line), ref_line)) {
-      # find the minimum x axis value from the plot
-      min_year <- min <- ggplot2::ggplot_build(plt)[["data"]][[2]] |>
-        as.data.frame() |>
-        dplyr::pull(x) |>
-        min() |>
-        round(digits = 2)
-      # find the reference point value for unfished
-      ref_point <- calculate_reference_point(
-        dat = stockplotr::example_data,
-        reference_name = "biomass_unfished"
-      ) / scale_amount
-      # add point to plot and add theme
-      final <- plt +
-        ggplot2::geom_point(ggplot2::aes(x = min_year - 1, y = ref_point)) + # should I keep -1 or set as first year?
-        theme_noaa()
-    } else {
-      final <- reference_line(
-        plot = plt,
-        dat = rp_dat,
-        label_name = "biomass",
-        reference = ref_line,
-        scale_amount = scale_amount
-      ) +
-        theme_noaa()
+    plt2 <- plt
+    # Check if length of ref_line = dat
+    # replicate value if not
+    if (length(ref_line) != length(dat)) ref_line <- rep(ref_line, length(dat))
+    # Put into for loop and add lines sequentially to plt
+    for (i in 1:length(ref_line)) {
+      # find the reference point value
+      if (is.null(names(ref_line[i]))) {
+        ref_line_x <- calculate_reference_point(
+          dat = dat[[i]],
+          reference_name = glue::glue("biomass_{ref_line[i]}"),
+          lbs = lbs
+        ) / scale_amount
+        ref_line_x <- setNames(ref_line_x, ref_line[i])
+      } else {
+        ref_line_x <- ref_line[i] / scale_amount
+      }
+      
+      if ("unfished" %in% names(ref_line_x)) {
+        # find the minimum x axis value from the plot
+        min_year <- min <- ggplot2::ggplot_build(plt2)[["data"]][[2]] |> # I think this was causing issues on linux?
+          as.data.frame() |>
+          dplyr::pull(x) |>
+          min() |>
+          round(digits = 2)
+        # add point to plot and add theme
+        plt2 <- plt2 +
+          ggplot2::geom_point(ggplot2::aes(x = min_year - 1, y = ref_point)) + # should I keep -1 or set as first year?
+          theme_noaa()
+      } else {
+        # add apply/purrr/or for loop for reference lines -- not just the first anymore
+        plt2 <- plt2 +
+          reference_line(
+            # conditionally add label name
+            label_name = ifelse(length(names(ref_line)) == 1, "biomass", names(dat)[i]), #"spawning_biomass",
+            ref_line = ref_line_x,
+            scale_amount = 1
+          )
+      }
     }
+    final <- plt2 + theme_noaa()
   }
 
   ### Make RDA ----
--- a/R/plot_biomass.R
+++ b/R/plot_biomass.R
@@ -156,7 +156,17 @@ plot_biomass <- function(
   }
   
   if (!is.data.frame(dat)) {
-    final <- final + ggplot2::scale_color_discrete(breaks = names(dat))
+    if (!is.null(names(ref_line))) {
+      if (any(
+        grepl(
+          glue::glue("^biomass_{ref_line}"),
+          # TODO: make this check more efficient
+          ifelse(is.data.frame(dat), dat[["label"]], sapply(dat, `[[`, "label"))
+        )
+      )) {
+        final <- final + ggplot2::scale_color_discrete(breaks = names(dat))
+      }
+    }
   }
 
   ### Make RDA ----
--- a/R/plot_biomass.R
+++ b/R/plot_biomass.R
@@ -141,58 +141,19 @@ plot_biomass <- function(
     ...
   )
   # Add reference line
-  # Conditions for ref line
-  # 1. all are comparing same value = msy, target, unfished...
-  # 2. custom input vector = c("sable23_msy"=30, "sable25_msy"=40) -- user must indicate which model in label otherwise it will assign in order of dat
-  # 3. input vector of labels = c("msy", "target")
-  # getting data set - an ifelse statement in the fxn wasn't working
   if (relative) {
     # don't add any reference line here and just add theme for final plot
     final <- plt + theme_noaa()
   } else {
-    plt2 <- plt
-    if (is.null(names(ref_line[i]))) {
-      # Check if length of ref_line = dat
-      # replicate value if not
-      if (length(ref_line) != length(dat)) ref_line <- rep(ref_line, length(dat))
-      # Put into for loop and add lines sequentially to plt
-      for (i in 1:length(ref_line)) {
-        # find the reference point value
-        if (is.null(names(ref_line[i]))) {
-          ref_line_x <- calculate_reference_point(
-            dat = dat[[i]],
-            reference_name = glue::glue("biomass_{ref_line[i]}"),
-            lbs = lbs
-          ) / scale_amount
-          ref_line_x <- setNames(ref_line_x, ref_line[i])
-        } else {
-          ref_line_x <- ref_line[i] / scale_amount
-        }
-        
-        if ("unfished" %in% names(ref_line_x)) {
-          # find the minimum x axis value from the plot
-          min_year <- min <- ggplot2::ggplot_build(plt2)[["data"]][[2]] |> # I think this was causing issues on linux?
-            as.data.frame() |>
-            dplyr::pull(x) |>
-            min() |>
-            round(digits = 2)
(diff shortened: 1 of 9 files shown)


R/plot_fishing_mortality.R
```diff
@@ -76,14 +76,19 @@ plot_fishing_mortality <- function(
...
-    label_name = "fishing_mortality",
-    reference = ref_line,
-    scale_amount = 1
-  ) + theme_noaa()
+  # Add reference line
+  # Conditions for ref line
+  # 1. all are comparing same value = msy, target, unfished...
+  # 2. custom input vector = c("sable23_msy"=30, "sable25_msy"=40) -- user must indicate which model in label otherwise it will assign in order of dat
+  # 3. input vector of labels = c("msy", "target")
+  # getting data set - an ifelse statement in the fxn wasn't working
```
sbreitbart-NOAA:
```suggestion
```
(this is repeated in utils_plot.R)

R/plot_fishing_mortality.R
```diff
@@ -76,14 +76,19 @@ plot_fishing_mortality <- function(
...
   )
-  # Add reference line and theme
-  final <- reference_line(
-    plot = plt,
-    dat = dat,
-    label_name = "fishing_mortality",
-    reference = ref_line,
-    scale_amount = 1
-  ) + theme_noaa()
+  # Add reference line
```
sbreitbart-NOAA, line 79: I think this information should be added to the `ref_line` documentation so that people know what to expect when adding multiple reference lines. Adding an example with multiple models and reference lines would help clarify these rules too, as would adding a section to the FAQs about how to add reference lines in a few different situations

R/utils_plot.R
```diff
@@ -529,11 +505,9 @@ cohort_line <- function(
 #' Preformatted reference line
 #'
 #' @inheritParams plot_spawning_biomass
-#' @param plot Plot object. A ggplot2 object where the reference line will be added
-#' @param dat Data frame. Standard data frame where reference point should be extracted
 #' @param label_name String. Name of the quantity that users want to
 #' extract the reference point from
-#' @param reference String of the reference point
+#' @param ref_line String of the reference point
```
sbreitbart-NOAA:
```suggestion
#' @param ref_line String. Reference point(s)
```

R/plot_spawning_biomass.R
```diff
@@ -202,41 +196,23 @@ plot_spawning_biomass <- function(
     ylab = spawning_biomass_label,
     group = group,
     # add check in case facet is returned as character(0)
-    facet = if (length(facet) > 0) facet else NULL,
-    ...
+    facet = if (length(facet) > 0) facet else NULL#,
```
sbreitbart-NOAA: Are the comments leftover from testing?

R/utils_plot.R
```diff
@@ -546,49 +520,39 @@ cohort_line <- function(
...
 #' }
 reference_line <- function(
-  plot,
-  dat,
   label_name,
-  reference,
+  ref_line,
   scale_amount = 1,
-  lbs = FALSE
+  model_name = "1"
```
sbreitbart-NOAA, line 553: I don't think this arg is documented. Will be helpful because the default is not immediately intuitive

R/utils_plot.R
```diff
@@ -1062,3 +1026,77 @@ plot_obsvpred <- function(
...
+    dat,
+    ref_line,
+    label,
+    lbs = FALSE,
+    scale_amount = 1
+) {
+  # Add reference line
+  # Conditions for ref line
+  # 1. all are comparing same value = msy, target, unfished...
+  # 2. custom input vector = c("sable23_msy"=30, "sable25_msy"=40) -- user must indicate which model in label otherwise it will assign in order of dat
```
sbreitbart-NOAA, line 1106: This is a vector while the input for dat must be a list. Can we choose one format for both for simplicity?

R/utils_plot.R
```diff
@@ -546,49 +520,39 @@ cohort_line <- function(
...
+    ),
+    # add line annotation
+    ggplot2::annotate(
+      geom = "text",
+      # TODO: need to change this for general process
+      x = -Inf,
+      y = ref_line_val / scale_amount,
+      label = glue::glue("{reference}"), # glue::glue("{stringr::str_replace_all(label_name, '_', '~')}[{reference}]"), # list(bquote(label_name[.(reference)])),
+      parse = TRUE,
+      hjust = -0.05, # slight offset so text doesn't hit border
```
sbreitbart-NOAA, line 578: Can you try to jitter the labels or else somehow separate them (maybe the x value gets 10% higher for each ref line or something)? My first example has the ref line labels on top of one another

R/utils_plot.R
```diff
@@ -119,16 +108,15 @@ plot_timeseries <- function(
               }
             }
           ),
-          alpha = 0.3
+          alpha = 0.3 #,
+          # show.legend = ifelse(all(is.na(dat$estimate_lower)), FALSE, TRUE)
```
sbreitbart-NOAA, line 112: This may be unrelated to this PR: leaving the geom as line works well, but specifying point gives a busy legend:

`plot_biomass(dat = list("petrale" = dat, "skate" = skate), ref_line = c("petrale_ref"=30000, "skate_ref"=400), geom = "point", module = "TIME_SERIES")`

Image

R/plot_biomass.R
```diff
@@ -141,37 +141,18 @@ plot_biomass <- function(
...
-      final <- plt +
-        ggplot2::geom_point(ggplot2::aes(x = min_year - 1, y = ref_point)) + # should I keep -1 or set as first year?
-        theme_noaa()
-    } else {
-      final <- reference_line(
-        plot = plt,
-        dat = rp_dat,
-        label_name = "biomass",
-        reference = ref_line,
+    final <- plt +
```
sbreitbart-NOAA, line 148: I like the messages shown when a ref line isn't found. Here, it shows up for one model but not both, even though zero ref lines are shown (i.e., no message warning about a lack of ref line for the stockplotr example data, but it's not shown):

`plot_biomass(dat = list("petrale" = dat, "skate" = skate), module = "TIME_SERIES")`
! biomass_msy not found for skate
Warning messages:
1: Removed 1 row containing missing values or values
outside the scale range (`geom_hline()`). 
2: Removed 1 row containing missing values or values
outside the scale range (`geom_text()`). 

Image

R/utils_plot.R
```diff
@@ -546,49 +520,39 @@ cohort_line <- function(
...
-  if (is.null(ref_line_val)) {
-    plot
-  } else {
-    plot +
-      ggplot2::geom_hline(
-        # ggplot2::aes(
+  list(
+    # Add geom for ref line
+    ggplot2::geom_hline(
+      data = data.frame(yintercept = ref_line_val, model = model_name, group_var = "1"),
```
sbreitbart-NOAA, line 562: Is this where the ".1" in the legend comes from? Is there a way to remove that?

Ex. 

Image

sbreitbart-NOAA (commented):
This feature is going to be so useful and important! Also great to see repeated code removed in favor of making a streamlined function useful for these and future plots.

I have some questions and suggestions. This functions as expected in most ways, but I think there are a few tasks left to do- like making sure this works and is documented clearly for various situations.

R/plot_fishing_mortality.R
Schiano-NOAA: Good idea! I will add this info into the documentation. In regards to the example, while I think this would be great, I hesitate to add this because we would ideally (1) need another example dataset and (2) need output from the same stock from different assessment cycles. While both is possible, adding another dataset could increase the size of the package a lot.

R/plot_spawning_biomass.R
Schiano-NOAA: Yes! I will clean up before merging. There is also commented out ... in plot_spawning_biomass (I believe) from when I was testing

R/utils_plot.R
Schiano-NOAA: I don't think this function has any documentation. Not sure if should be exported or left in utils?

Edit: I see that this one is exported. I will update then

R/utils_plot.R
Schiano-NOAA: I think we have to pick list if we do this. I don't think data frames can be in a named vector. If others have input on this, it would be helpful

R/plot_fishing_mortality.R
sbreitbart-NOAA: How about just putting `\dontrun{}` brackets around that example? Getting the plot would be awesome but simply seeing how one would set up the function command would still be very helpful

R/utils_plot.R
Schiano-NOAA: On to keep as is or swtich. Keep in mind that authors in asar is a named vector along with all other approaches to arguments in both packages that can take multiple inputs

R/utils_plot.R
Schiano-NOAA: I don't know how jittering would be helpful because then you probably wouldn't be able to tell which label is associated with which line when they are close together. I am not sure what to do about this. A potential other option is to remove the label and see if I can add the ref line label in the legend, but I am already having so much trouble with the legends.

R/utils_plot.R
Schiano-NOAA: All options have busy legends and I am unable to fix it. I have spent most of this development time playing around with it. Maybe after we merge this, you can try and tackle this issue and add it as a fix?

R/plot_biomass.R
Schiano-NOAA: That's interesting because it should show for both if both don't appear. It's in a loop

R/utils_plot.R
Schiano-NOAA: No it is an automatically generated label from `ggplot` because we use `color = interaction(model, group_var)`. I have conditions set to stop this when length of model = 1, but they aren't working. Again, this is the same issue I mentioned above about the busy legend and I can not figure out how to change it.

R/plot_fishing_mortality.R
Schiano-NOAA: Oh yeah duh 🤦 that can be added!

R/utils_plot.R
sbreitbart-NOAA: Ok- if there's already a mix of input types for args, I'm alright keeping this one as-is

R/utils_plot.R
sbreitbart-NOAA: That's fair (re: jittering). Ultimately, I think having the ref line labels be in the legend would be best for avoiding overlapping and general clarity of the plot (fewer elements inside it). I hear you re: how challenging the legends have been to set up. I'll make this a task in the project so we can address it in the future if we decide

R/utils_plot.R
sbreitbart-NOAA: I'm sure you've tried a bunch of ideas to overcome it. If all else fails, we can easily share how to manually adjust the plot's legend (example: this gets rid of the "skate.1" and "petrale.1" legend items):

> plot_biomass(dat = list("petrale" = dat, "skate" = skate), ref_line = c("petrale_ref"=30000, "skate_ref"=400), geom = "point", module = "TIME_SERIES") **+ scale_color_discrete(breaks = c("petrale", "skate"))**

image

Schiano-NOAA: Hmm maybe I can use that in the code? I can try it out and if it doesn't work then we can open an issue

R/plot_biomass.R
Schiano-NOAA: This is fixed now! Good catch

Schiano-NOAA:
@sbreitbart-NOAA Tysm for your review! I addressed all you comments/concerns. Was there anything else to mention before I merge?

sbreitbart-NOAA:
> @sbreitbart-NOAA Tysm for your review! I addressed all you comments/concerns. Was there anything else to mention before I merge?

Super! Not that I can think of 👍

PR #316, opened 2026-07-28
[feat]: plot_discard

### LINKED ISSUE #266: [feature] plot_discard
Model fit to discards (AM): if applicable

Schiano-NOAA:
A new plot for {stockplotr}. The discards of the stock over time found in the data. Follows the same conventions and features as other plots.

Note: not all output contains discards. Also, Rceattle does not explicitly have discards, so this is not compatible.


Schiano-NOAA:
@sbreitbart-NOAA this is now ready for review. Please note that the current example_data is not going to work since I had to make updates to the converter. I am not going to do this manually as the gh action should do this after merge, but I will see if it works. Also, not every dataset contains discards explicit in the model, so if it's not there, it is not unexpected. Rceattle does not export discards so it does not work with this function.

R/plot_discard.R
```diff
@@ -0,0 +1,154 @@
...
+#' All plotting functions automatically recognize indexing variables and will
+#' use them in groupings and/or facetting. @seealso [process_data()].
+#'
+#' @seealso [convert_output()], [plot_timeseries()], [calculate_reference_point()], [reference_line()], [filter_data()], [process_data()], [export_kqs()], [insert_kqs()], [create_rda()]
+#' 
+#' @export
+#'
+#' @examples
+#' plot_discard(
+#'   dat = example_data,
```
sbreitbart-NOAA:
```suggestion
#'   dat = stockplotr:::example_data,
```

R/plot_discard.R
```diff
@@ -0,0 +1,154 @@
...
+      facet = if (length(facet) > 0) facet else NULL,
+      group = group
+    )
+  }
+  
+  ### Make RDA ----
+  if (make_rda) {
+    # Obtain relevant key quantities for captions/alt text
+    mod.fit.discards.min <- min(discards$estimate) |> round(digits = 3)
+    mod.fit.discards.max <- max(discards$estimate) |> round(digits = 3)
```
sbreitbart-NOAA: These don't work for BAM (output NA). They do work for SS3

R/plot_discard.R
```diff
@@ -0,0 +1,154 @@
...
+    prepared_data,
+    group = group,
+    facet = facet,
+    lbs = lbs
+  )
+  discards <- p_dat[[1]]
+  group <- p_dat[[2]]
+  facet <- p_dat[[3]]
+  
+  # make the plot
```
sbreitbart-NOAA, line 91: The legend titles (fleet, model) are lowercase (pls capitalize): 

e.g., with the example_data:
Image

sbreitbart-NOAA (commented):
This looks great! I only had a couple minor comments. I think this should be ready to merge soon!

sbreitbart-NOAA:
One more suggestion: it would be great to have a message if discards aren't in the model, so the user gets a clear reason why the plot isn't working rather than a potentially confusing error msg

Schiano-NOAA:
@sbreitbart-NOAA do you think that the message stating that discards was not found is not sufficient? I think alternatively if they know their model, would they even try to use plot_discard?

sbreitbart-NOAA:
> @sbreitbart-NOAA do you think that the message stating that discards was not found is not sufficient? I think alternatively if they know their model, would they even try to use plot_discard?

Oh, those are good points- I wasn't thinking of either. Proceed!

PR #318, opened 2026-07-29
Adjust plots to handle log-transformed data

### LINKED ISSUE #141: plot_timeseries does not work for log scale
when using plot_timeseries, we set y bounds to include 0 so when things are in log scale then the plot is awkward

sbreitbart-NOAA:
* Adjust y axis limits so that they don't have to include zero if values are negative, as per https://github.com/nmfs-ost/stockplotr/issues/141
* Adjust timeseries and obsvpred functions to create area plots that work with the above change
* Instruct users how to show that values were log-transformed in the units, with messages

### DIFF
--- a/R/utils_plot.R
+++ b/R/utils_plot.R
@@ -91,7 +91,11 @@ plot_timeseries <- function(
             # TODO: add more groupings
             # shape = ifelse(any(grepl("shape", names(group))), .data[[group[[grep("shape", names(group))]]]], 1),
             # color = ifelse(any(grepl("color", names(group))), .data[[group[[grep("color", names(group))]]]], "black")
-            color = model,
+            color = if (length(unique(.data[["model"]])) > 1) {
+              interaction(model, group_var)
+            } else {
+              group_var
+            },
             shape = group_var
           ),
           # size = point_size,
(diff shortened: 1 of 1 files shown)


R/utils_plot.R
```diff
@@ -91,7 +91,11 @@ plot_timeseries <- function(
             # TODO: add more groupings
             # shape = ifelse(any(grepl("shape", names(group))), .data[[group[[grep("shape", names(group))]]]], 1),
             # color = ifelse(any(grepl("color", names(group))), .data[[group[[grep("color", names(group))]]]], "black")
-            color = model,
+            color = if (length(unique(.data[["model"]])) > 1) {
```
Schiano-NOAA, line 94: Is this not already in main?

Schiano-NOAA (approved):
LGTM! I had one minor (unrelated question)

R/utils_plot.R
sbreitbart-NOAA: Nope 👀  https://github.com/nmfs-ost/stockplotr/blob/70245d1c3fab4e97932722fa5d5e8ca9684166dd/R/utils_plot.R#L94

sbreitbart-NOAA:
> LGTM! I had one minor (unrelated question)

Thanks!

PR #336, opened 2026-08-10
[fix]: Fix rda-exporting mechanism for tables

sbreitbart-NOAA:
- Fix component of table being exported to rda
- update tests to ensure tables exported properly


Schiano-NOAA (approved):
LGTM!

PR #337, opened 2026-08-10
Move kableExtra from Imports to Suggests

### LINKED ISSUE #325: [chore]: Move kableExtra from imports to suggests
`kableExtra` is not a dependency of `stockplotr`, but the package is used for tests and to show that users can use `add_theme()` for a `kableExtra` table. 

The package needs to be moved from imports to suggests in the [DESCRIPTION file](https://github.com/nmfs-ost/stockplotr/blob/main/DESCRIPTION).

Melissa-Karp:
Moved the kableExtra from imports to suggests in the DESCRIPTON filet address issue #325

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,6 @@ Imports:
     ggplot2 (>= 4.0.0),
     glue,
     gt,
-    kableExtra,
     purrr,
     rlang,
     scales,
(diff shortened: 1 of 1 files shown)


Schiano-NOAA (approved):
LGTM (looks good to me)! Thank you for helping out with this 😄

Issue #117, opened 2025-07-10
Find package improvements with `goodpractice` package


sbreitbart-NOAA:
The `goodpractice` package gives "advice about good practices when building R packages. Advice includes functions and syntax to avoid, package structure, code complexity, code formatting, etc."

**Steps:**
1. Install the [`goodpractice`](https://github.com/ropensci-review-tools/goodpractice/) and stockplotr R packages
```
install.packages("pak")
pak::pak("goodpractice")
pak::pak("nmfs-ost/stockplotr")
```
2. Find where the `stockplotr` package is saved on your machine. This workflow should work:
```
# Find where your packages live
pkg_location <- .libPaths()
# Then find where stockplotr lives, specifically. It might be:
stockplotr_location <- fs::path(pkg_location, "stockplotr")
```
3. Run `goodpractice`
```
goodpractice::gp("stockplotr_location")
```
4. Address the package's advice 🙂

sbreitbart-NOAA:
Now, the `goodpractice` package has a [guide](https://github.com/ropensci-review-tools/goodpractice/blob/main/inst/skills/goodpractice4agents.md) that instructs AI agents how to review the package. Another way to address this issue is to share this guide with an agent and ask it to review the package based on the guide, then you review the resultant changes. When you've reviewed and edited the agent's changes, you can open a pull request for stockplotr.

e-perl-NOAA:
Some of this has been implemented but the following list of things to be addressed is what I produced, AI agents were less helpful than expected - the branch is [`docs-goodpractice`](https://github.com/nmfs-ost/stockplotr/tree/docs-goodpractice). This is not a hard issue but it does take a long time, so I would recommend putting that somewhere either in a category in an attached project board or as a label.

```
Scanning roxygen comments in R/ files...

Possible missing @noRd for internal helper:
  File: R/afsc_risk_table.R 
  Function: # risk_table <- function(dat, end_year, tier, c1, c2, c3) { 

Possible missing @param docs:
  File: R/afsc_risk_table.R 
  Function: # risk_table <- function(dat, end_year, tier, c1, c2, c3) { 
  Missing: dat, end_year, tier, c1, c2, c3) { #   # for cells in dat with numeric values, format the numbers #   # with built-in comma function #   for (column in names(dat)) { #     if (is.numeric(dat[[column]])) { #       dat[[column]] <- sapply(dat[[column]], comma 

Possible missing @noRd for internal helper:
  File: R/convert_output.R 
[...]

PR #341, opened 2026-08-10
Some elements of good practices implemented

e-perl-NOAA:
This is not finished - see notes in issue #117

### DIFF
--- a/R/convert_output.R
+++ b/R/convert_output.R
@@ -623,19 +623,6 @@ convert_output <- function(
                     df4 <- df4 |>
                       dplyr::filter(!(label %in% err_names[2:length(err_names)]))
                   }
-                  # Find overlapping error values if still present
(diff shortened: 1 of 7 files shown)


R/utils_plot.R
Schiano-NOAA, line 1: Note: the filter_data should not be showing errors and have missing argument documentation because it is sharing the same documentation from spawning_biomass through inheritParams. I am curious why the check said this was missing

R/convert_output.R
Schiano-NOAA, line 1: need to check if this is moved to utils. If so, can be removed, if not then this needs to stay. It is referenced in a couple lines below

Schiano-NOAA (changes requested):
I will make the changes neccessary to merge this. Thanks @e-perl-NOAA for getting this started. This is a big help

PR #350, opened 2026-08-20
[feat] catch table for exec summary

sbreitbart-NOAA:
Table showing catch, for the executive summary

@Schiano-NOAA I didn't use the typical process_table() step because it didn't seem to fit with this pipeline. Also made this suitable for creating one table at a time


inst/resources/captions_alt_text_template.csv
```diff
@@ -32,7 +32,7 @@ proj.catch,figure,Forecasted catch in proj.catch.units over future years for dif
 proj.biomass,table,Forecasted biomass over future years for different fishing mortality scenarios.,
 report.version.model.changes,table,Document version history briefly describing when major changes or updates are made to each version of the report.,
 projections,table,"Forecasted catch, spawning biomass, and fishing mortality over future years. ",
-catch,table,Observed (points) and model estimated (line) catch over time for fleet catch.fleet.,
+catch,table,Total catch over time.,
```
Schiano-NOAA:
```suggestion
tot.catch,table,Total catch over time.,
```

R/table_catch.R
```diff
@@ -0,0 +1,184 @@
...
+  unit_label <- label_magnitude(
+    label = "",
+    unit_label = unit_label,
+    scale_amount = scale_amount
+  )
+
+  # Filter data for catch
+  prepared_data <- filter_data(
+    dat = dat,
+    label_name = "catch",
```
Schiano-NOAA:
```suggestion
    label_name = "^catch",
```

Do you know if there's any reason we can't do this? I think it would reduce the number of module options in the first step

inst/resources/captions_alt_text_template.csv
Schiano-NOAA: After testing, I noticed this table is showing the results by fleet, so I think we need to include that still?

tests/testthat/test-table_index.R
Schiano-NOAA, line 1: Why do you need to do this?

tests/testthat/test-table_landings.R
Schiano-NOAA, line 1: same here?

Schiano-NOAA (changes requested):
I think this overall is great! I made a few minor suggestions. Could you please change the table name to "table_total_catch" to add clarity. This seems to work as expected on testing so glad for that! There was some inconsistency in the caption vs what was returned which I mentioned in a comment. Once these things are changed and lables/test/everything else are aligned with these edits, then should be good to merge

tests/testthat/test-table_index.R
sbreitbart-NOAA: The test name was duplicated (and inaccurate) for the second "test_that("rda file made when indicated", {" line in both functions, and the code in that second duplicated test fit into the first rda-related testing block, so I thought it would better organize the code. Didn't change any of the tests- just the way they're presented

tests/testthat/test-table_landings.R
sbreitbart-NOAA: same as above

R/table_catch.R
sbreitbart-NOAA: Tested it with several examples and all work, so seems fine to me!

sbreitbart-NOAA:
> I think this overall is great! I made a few minor suggestions. Could you please change the table name to "table_total_catch" to add clarity. This seems to work as expected on testing so glad for that! There was some inconsistency in the caption vs what was returned which I mentioned in a comment. Once these things are changed and lables/test/everything else are aligned with these edits, then should be good to merge

Thanks Sam! Made the changes you requested and gave some more context about the test-related changes. If you're on board, I'll merge when you're ready

Schiano-NOAA:
Perfect! Thanks for doing that and go ahead with merging!

Schiano-NOAA (approved):

PR #356, opened 2026-08-31
[fix]: adjust how module_name is selected in `filter_data()` step

### LINKED ISSUE #355: Multiple module names are chosen if not provided during function call
The new https://github.com/nmfs-ost/stockplotr/pull/353 PR introduces a bug where the "preselected" module name, meant to be chosen only if the user doesn't choose one, is added even when the user interactively selects their own module name in the console. 

Also, there's a plus sign next to the first module name now.

Example:

This works:
`table_total_catch(dat = stockplotr::example_data, module = "TIME_SERIES")`

This doesn't, since it automatically adds "DERIVED_QUANTITIES" (the first module option) to the list of chosen module names, even if I choose "TIME_SERIES" in the console:
```
table_total_catch(stockplotr::example_data)
! Multiple module names found in data. 
Select one or more of the following module names

1: + DERIVED_QUANTITIES
2:   PARAMETERS
3:   CATCH
4:   CATCH_AT_AGE
5:   TIME_SERIES

Enter one or more numbers separated by spaces and then ENTER, or 0 to cancel
1: 5
Error in names(catch[i]) == "uncertainty" && is.na(unique(catch[[i]])) : 
  'length = 25' in coercion to 'logical(1)'
```

Schiano-NOAA:
- Remove preselect argument and set selection to default in an ifelse statement
- Resolves #355

### DIFF
--- a/R/utils_plot.R
+++ b/R/utils_plot.R
@@ -783,12 +783,16 @@ filter_data <- function(
       }
       if (interactive()) {
         if (interactive) {
-          question1 <- utils::select.list(
+          selected <- utils::select.list(
             options,
             multiple = TRUE,
-            preselect = options[1],
             title = "Select one or more of the following module names"
           )
+          question1 <- if (length(selected) == 0 || identical(selected, "")) {
+            options[1]
+          } else {
+            selected
+          }
           # use <<- to export module to environment for use with key quantity calc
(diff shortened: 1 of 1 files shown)


sbreitbart-NOAA (approved):
Thanks for the quick fix!

PR #365, opened 2026-09-04
[feat] Extract and format data for asar to SIS function

sbreitbart-NOAA:
Extract and format data for asar to SIS function
Note: this branch is a companion to https://github.com/nmfs-ost/asar/tree/sis-data, which intakes the templates (blank or filled in, with this function) and sends them to Google Drive

@Schiano-NOAA This works by extracting data from rdas, not creating them within the function. I'm not sure that it'd be worth adding in the capability to create rdas within the function, as it'd need to be highly interactive (asking authors for arguments that impact key quantities), and this works even if rdas are missing.

This still needs some work (see below) but wanted to share it/ask for feedback/spark discussion before I got too far:

- [ ] unit tests
- [ ] documentation: what the output should look like, which rdas are expected (and how to export them via the functions), etc. (in a vignette?)


R/extract_sis_data.R
```diff
@@ -0,0 +1,385 @@
+#' Extract data from model results to send to SIS
+#'
+#' Semi-automate the extraction of key quantities from model results for eventual transmittance to SIS via `asar::export_to_sis()`.
+#' 
+#' @param sis_data_dir Path. Location of the existing sis_assmt_template.csv
```
Schiano-NOAA: I think it would improve clarity here if we assumed absence first then said if present. For example, the location to place the sis template or if present the path to the sis_template

R/extract_sis_data.R
```diff
@@ -0,0 +1,385 @@
...
+#'
+#' Semi-automate the extraction of key quantities from model results for eventual transmittance to SIS via `asar::export_to_sis()`.
+#' 
+#' @param sis_data_dir Path. Location of the existing sis_assmt_template.csv 
+#' file, or, if absent, where a new version of sis_assmt_template.csv and
+#' sis_ts_template.csv should be saved.
+#' 
+#' Default: The working directory.
+#' 
+#' @param kq_dir Path. Location of the existing key_quantities.csv file, or, if absent, where a new version of the file should be saved.
```
Schiano-NOAA: same here

R/extract_sis_data.R
```diff
@@ -0,0 +1,385 @@
...
+#' Default: The working directory.
+#' 
+#' 
+#' @param figures_tables_dir Path. Location of the existing 'figures' and 'tables' directories.
+#' 
+#' Default: The working directory.
+#' 
+#' @details This function acts within the following workflow:
+#' 
+#' 1. When a stock assessment is scheduled to conclude, SIS will generate an
```
Schiano-NOAA, line 21: What do you mean SIS will generate? I think clarity here is needed saying a specific function.

R/extract_sis_data.R
```diff
@@ -0,0 +1,385 @@
...
+#'  attachment or prompt containing metadata and identifiers.
+#' 2. The user will open two csv files containing placeholders for all of the data required by SIS: sis_assmt_template.csv (assessment summary data) and sis_ts_template.csv (time series data). There are three ways to obtain these files:
+#' 2a. Generate the files by running `asar::create_blank_sis()`
+#' 2b. Locate the files in the "report" folder generated by running `asar::create_template()`.
+#' 2c. Run `stockplotr::extract_sis_data()`, which will populate the templates with data originating from a converted model results file.
```
Schiano-NOAA: I would make this the first option as this is in the documentation for this function. Also mention that this is done automatically. I am not sure it's clear that they don't have to create the blank files first.

R/extract_sis_data.R
```diff
@@ -0,0 +1,385 @@
...
+#' 3. The user will add the remaining necessary data into the csv files, ensuring that all required fields are completed.
```
Schiano-NOAA, line 27: Is this output in the console for what is missing? This also gives me an idea that maybe we can add some indicator or placeholder in missing parts so when this gets sent to SIS or the google drive, maybe the function can do a check for missing values and ask the user if they are sure they want to send a file with missing fields.

R/extract_sis_data.R
```diff
@@ -0,0 +1,385 @@
...
+#' 4. Run `export_to_sis()`, which will format and upload this data to a specific Google Drive folder.
```
Schiano-NOAA, line 28: Maybe we should make a wrapper like `add_accessibility` that performs this fxn and `export_to_sis` together

R/extract_sis_data.R
```diff
@@ -0,0 +1,385 @@
...
+    }
+  
+  if (length(primary) == 0) {primary <- NA} 
+  
+  category_pairs <- list(
+    "fishing_mortality" = "Fmort",
+    "recruitment" = "Recruitment",
+    "catch" = "Catch",
+    "spawning_biomass" = "Spawners",
+    "abundance" = "Abundance", # aka biomass
```
Schiano-NOAA, line 320: abundance and biomass are not the same

Schiano-NOAA:
This is really coming together! I do feel quite strongly that the users should be able to use this function without creating the RDAs; however, I do see why the approach here is more reasonable. Playing devil's advocate, I do think that if a user wants to send their info to SIS and doesn't use `stockplotr` for their figures and tables, they will need to make all of the necessary ones before running this function, which is just as highly interactive as having this function operate without rdas.

I think we need to really think about the purpose of this function. Is it only for users who use `stockplotr` in its entirety or is it for people to pick and choose what they want to do even if it's simply converting their output and easily sending their results to SIS?

I can see both sides to this. Since `stockplotr` isn't in v1.0 quite yet and not planned to get there until end of June 2027, adoption by a large potion of the community wouldn't be expected for sometime after that. I think this approach to separating functions out though will be a big help to the transition for SIS and the user base.

With that said and thinking on it, proceed as planned and continue with this approach. I think we can assess the utility of it with some test cases (the steering committee or other willing users) once we get this out there and talk with the SIS devs.

R/extract_sis_data.R
sbreitbart-NOAA: If the user doesn't have the necessary rdas, there will be messages about how "Abundance data was not extracted from the 'abundance_at_age_figure.rda' file." (etc).
I think it's a great idea to add some more checks for missing values to `export_to_sis()`! I'll add it as a TODO

R/extract_sis_data.R
sbreitbart-NOAA: How about "automatically assign"?

R/extract_sis_data.R
sbreitbart-NOAA: I like the idea of saving the user time, but that might come at the cost of them not inspecting their files before sending to SIS. I think we should encourage them to make sure it looks right before doing so

R/extract_sis_data.R
sbreitbart-NOAA: I was confused too; Jeff confirmed that "Abundance = Biomass - or - Spawners". I'm going to inquire further

PR #371, opened 2026-09-16
[docs]: add acronym exceptions

Schiano-NOAA:
Adding a new section describing and defining the acronyms to the exceptions in the naming conventions section.


Schiano-NOAA:
@kellijohnson-NOAA here are a few potential exceptions to make, but honestly was less than I thought when going through the list. Are there any additional ones you can think of?

